Issue #13, opened 2017-08-03
Move from glide to dep


pires:
See https://github.com/golang/dep

PR #14, opened 2017-08-04
Switch from 'glide' to 'dep'.

brunomcustodio:
This is my attempt at #13. Here's what's going on:

- From a fresh clean `master`, I ran `dep init`.
- Removed `vendor/` ignores from both `.{docker,git}ignore`.
- Added the `vendor/` folder to Git.
- Edited the Dockerfile so that Glide isn't required anymore.
- Removed `glide.{lock,yaml}`.
- Checked it builds both outside and inside the container.

IDK for sure how this will fit with the `rework` branch, but I believe it won't be hard to "rebase" (as in, manually re-apply changes) that branch on top of `master` once this is merged.


pires:
I understand this seems the easiest step, but it's not. First, the imports need to be reworked to use client-go and maybe other dependencies have changed as well. Only then `dep init`. I'm owning this for now so don't waste more time with it.

PR #15, opened 2017-08-08
To be merged into 'rework'.

brunomcustodio:
@pires I didn't want to step into `rework` directly, hence this PR. This would benefit from your knowledge and incisive eye.

The project builds, but I've not (yet) verified it "works". I'm dropping a few notes to help you through the review:

- I've introduced `hack/codegen.sh`, which pulls the `deepcopy-gen` generator, builds it into `bin/` and generates `pkg/spec/zz_generated.deepcopy.go` — this contains all the necessary `DeepCopy*` implementations.
- There were a few compilation errors (erroneous references, etc...), so I started fixing them.
- When the only error left was `resourcelock.New()` complaining about `kubernetes.Interface`, I found out that maybe it was time to switch from `k8s.io/client-go/kubernetes` to `k8s.io/client-go/typed/core/v1`, so I did that everywhere.
- Finally, I tweaked the `Dockerfile` because the build was failing in this branch — Alpine uses `musl`, and as such requires `CGO_ENABLED=0`. Also, I believe static compilation is not strictly necessary as we're building in an Alpine container to run in another Alpine container.


Dockerfile
```diff
@@ -1,11 +1,10 @@
...
 WORKDIR /go/src/github.com/pires/nats-operator
-RUN \
-	apk add --update git; \
-	go get -u github.com/golang/dep/cmd/dep; \
-	dep ensure; \
-	go build -ldflags '-w -extldflags=-static' -o /nats-operator ./cmd/operator/main.go
+RUN apk add --update git
+RUN go get -u github.com/golang/dep/cmd/dep
+RUN dep ensure
+RUN CGO_ENABLED=0 go build -installsuffix cgo -o /nats-operator ./cmd/operator/main.go
```
pires: Why split into four different steps?

pkg/spec/cluster.go
```diff
@@ -30,7 +30,7 @@ import (
 // TODO: supports object store like s3
 type StorageType string
```
pires: Please remove the above. I'm seeing a lot of artifacts that I find it weird they're still around. Seems my "grep" skills are too rusty.

pkg/spec/zz_generated.deepcopy.go
```diff
@@ -0,0 +1,348 @@
...
+
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new TLSPolicy.
+func (in *TLSPolicy) DeepCopy() *TLSPolicy {
+	if in == nil {
+		return nil
+	}
+	out := new(TLSPolicy)
+	in.DeepCopyInto(out)
+	return out
+}
```
pires: If the build always generates code, why version it?

Dockerfile
brunomcustodio: Mainly because this is an intermediate container — so we don't have to worry about layers and image size — and I believe it improves readability.

pkg/spec/cluster.go
brunomcustodio: Done :+1:

Dockerfile
pires: Meh OK.

pkg/spec/zz_generated.deepcopy.go
brunomcustodio: That's one thing I'd like to discuss with you. ATM the build doesn't generate code automatically. We have two options:

1) Run `hack/codegen.sh` manually after each change to `pkg/spec/*` and version the generated file (the current approach).
2) Add a `RUN ./hack/codegen.sh` step in the `Dockerfile` and ignore `zz_generated.*.go`.

Which one do you prefer? I was about to go with (2), in fact, but ended up going with (1) since the _codegen_ tool is evolving quickly and I feel it's somewhat safer to version what we're actually building. But I don't have a strong opinion, so please let me know which one do you prefer.

pires: I can second that.

PR #19, opened 2018-02-16
Cluster formation via A records and TLS support

wallyqs:
Hi. I have been looking into adding TLS support both between the
clients and the servers and routes to other servers.  Currently, the
cluster formation is using auto discovery but we may hit a roadblock
due to auto discovery using the resolved IPs of the hosts.

This branch is a wip still, but shows some of the cluster formation changes
which I'm thinking of in order to support TLS in the routes from a
NATS cluster.

Have gist here that shows how the approach works:
https://gist.github.com/wallyqs/696b81427df7c239fb34946eb1ae9f92

Basically, similar to the etcd-operator we will create an A record for
each one of the pods in the cluster under a wildcard certificate, then
the user would be able to get which is the secret to use for the
cluster via the manifest for the CRDT.

For example:

```
nslookup example-nats-cluster-1

Name:      example-nats-cluster-1
Address 1: 172.17.0.6 nats-707.example-nats-cluster-1.default.svc.cluster.local
Address 2: 172.17.0.5 nats-503.example-nats-cluster-1.default.svc.cluster.local
Address 3: 172.17.0.4 nats-753.example-nats-cluster-1.default.svc.cluster.local
```

Then the clients would use the non-headless service to connect to it:

```
nslookup example-nats-cluster-1-client

Name:      example-nats-cluster-1-client
Address 1: 10.103.185.136 example-nats-cluster-1-client.default.svc.cluster.local
```

Some related pending items that I think of:

- [x] Use volumes to mount the secrets with the certificates
- [x] Create config map with configuration shared between NATS pods
- [x] Modify to use stable names for the set of pods instead of auto generated
- [x] Use namespace of cluster as part of the A record subdmain

Let me know if you have some early feedback! Cheers.


wallyqs:
Closing in favor of #20

PR #22, opened 2018-03-07
Updates for v1beta1

wallyqs:
Rebased version of PR at https://github.com/nats-io/nats-operator/pull/20/files to include the fixed imports for the repo. Also includes:

- [X] Bump of version of API group and CRD to `nats.io/v1beta1`
- [X] Minikube based Travis build
- [X] README updates and docs
- [X] Updates to script for code generation

PR #23, opened 2018-03-07
Maintenance fixes

wallyqs:
Few maintenance related fixes:

- `gofmt` pass to the repo
- Update Deployment API to use for the operator
- Remove readiness probe which is not supported by official image

Part of bigger PR at: https://github.com/nats-io/nats-operator/pull/22/files


README.md
```diff
@@ -1,26 +1,94 @@
...
-**Project status: *ALPHA***
-
-The API, spec, status and other user facing objects are subject to change.
-We do not support backward-compatibility for the alpha releases.
+[k8s-home]: http://kubernetes.io
 
 ## Requirements
 
-- Kubernetes 1.7+
+- Kubernetes v1.7+
```
pires: Actually, after your [change here](https://github.com/nats-io/nats-operator/pull/23/files#diff-e87d9ff03678a45cd0da2aca55f6fe9cR1) this will only work on Kubernetes 1.8 or newer.

pkg/util/kubernetes/kubernetes.go
```diff
@@ -114,8 +114,7 @@ func CreateMgmtService(kubecli corev1client.CoreV1Interface, clusterName, cluste
 			Name:       "monitoring",
 			Port:       constants.MonitoringPort,
 			TargetPort: intstr.FromInt(constants.MonitoringPort),
-			Protocol:   v1.ProtocolTCP,
-		},
+			Protocol:   v1.ProtocolTCP,		},
```
pires: This formatting was made by gofmt or yourself? Nothing special but a little weird.

pkg/util/kubernetes/kubernetes.go
```diff
@@ -199,7 +198,6 @@ func NewNatsPodSpec(clusterName string, cs spec.ClusterSpec, owner metav1.OwnerR
 
 	container := natsPodContainer(clusterName, cs.Version)
 	container = containerWithLivenessProbe(container, natsLivenessProbe(cs.TLS.IsSecureClient()))
-	container = containerWithReadinessProbe(container, natsReadinessProbe(clusterName))
```
pires: So no readiness probe? This means any NATS pod, regardless of its state, will be resolvable and proxied to, no matter its state, which may result in bad client UX.

pkg/util/kubernetes/kubernetes.go
wallyqs: ah interesting, looks like it does that when there is no trailing comma

README.md
wallyqs: thanks, yes should bump in the readme then

pkg/util/kubernetes/kubernetes.go
wallyqs: looks like didn't do gofmt well here, addressed in other commit

pkg/util/kubernetes/kubernetes.go
wallyqs: right now yes, though we could revisit this part later and have something like a parameter in the server `/routez` to assert that the cluster is formed for the case when the pod is warming up.

PR #26, opened 2018-03-08
Use ConfigMap with the explicit routes for the cluster

wallyqs:
Besides the pair of services the created NatsCluster also has a config map that is used to include the full list of the non-failed pods so far, this as one step in order to be able to use wildcard certificates for the cluster routes.

```sh
$ kubectl get cm test-nats-cluster-1 -o yaml
apiVersion: v1
kind: ConfigMap
data:
  nats.conf: |2

      "port": 4222

      "http_port": 8222

      "cluster": {
        "port": 6222

        "routes": [
          "nats://nats-4dnrj8qh1j.nats-cluster-1-mgmt.default.svc:6222"

          "nats://nats-cvms254b0l.nats-cluster-1-mgmt.default.svc:6222"

          "nats://nats-w142n49wt6.nats-cluster-1-mgmt.default.svc:6222"
        ]
      }
```

### DIFF
--- a/pkg/cluster/reconcile.go
+++ b/pkg/cluster/reconcile.go
@@ -61,7 +61,7 @@ func (c *Cluster) reconcileSize(pods []*v1.Pod) error {
 		}
 		err = c.updateConfigMap()
 		if err != nil {
-			c.logger.Warningf("error update the shared config map: %s", err)
+			c.logger.Warningf("error updating the shared config map: %s", err)
 		}
 
 	} else if currentClusterSize > spec.Size {
--- a/pkg/cluster/reconcile.go
+++ b/pkg/cluster/reconcile.go
@@ -26,7 +26,7 @@ import (
 // - if the cluster needs upgrade, it tries to upgrade existing peers, one by one.
 func (c *Cluster) reconcile(pods []*v1.Pod) error {
 	c.logger.Debugln("Start reconciling...")
-	defer c.logger.Infoln("Finish reconciling")
+	defer c.logger.Debugln("Finish reconciling")
 
 	spec := c.cluster.Spec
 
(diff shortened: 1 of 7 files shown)


pkg/cluster/reconcile.go
```diff
@@ -50,14 +50,20 @@ func (c *Cluster) reconcile(pods []*v1.Pod) error {
...
 	if currentClusterSize < spec.Size {
 		c.status.AppendScalingUpCondition(currentClusterSize, c.cluster.Spec.Size)
-		if err := c.createPod(); err != nil {
+		_, err := c.createPod()
+		if err != nil {
 			return err
 		}
+		err = c.updateConfigMap()
+		if err != nil {
+			c.logger.Warningf("error update the shared config map: %s", err)
```
ColinSullivan1: typo, "updating"

pkg/conf/natsconf.go
```diff
@@ -0,0 +1,73 @@
+// natsconf is a package for producing NATS config programmatically.
+package natsconf
```
ColinSullivan1: Long term, we'll want to find a way to more closely integrate with NATS server, but certainly OK for now.

wallyqs: Agreed, should find a better way here eventually

pires:
Given the example above, what is `"nats://nats-4dnrj8qh1j.nats-cluster-1-mgmt.default.svc:6222`? One service per pod? 🤔

wallyqs:
In that example, there would be a single service named `nats-cluster-1-mgmt` and then in the PodSpec the subdomain it is mapped so that it is the same as the one from the service, plus the name of the pod as the hostname: 
https://github.com/nats-io/nats-operator/blob/4efdeeae68c848b1c86f3260c6f588b2e098de7c/pkg/util/kubernetes/kubernetes.go#L333-L335

Using the hostname+subdomain and headless service in combination creates the additional A record which can also be used to connect to the node, doing nslookup would result in something like:

```
$ nslookup nats-cluster-1-mgmt.default.svc

Name:     nats-cluster-1-mgmt.default.svc
Address 1: 172.17.0.8 nats-4dnrj8qh1j.nats-cluster-1-mgmt.default.svc.cluster.local
Address 2: 172.17.0.7 nats-cvms254b0l.nats-cluster-1-mgmt.default.svc.cluster.local
Address 3: 172.17.0.6 nats-w142n49wt6.nats-cluster-1-mgmt.default.svc.cluster.local
```

pires:
I don't follow totally but if it works, great.

PR #27, opened 2018-03-09
Add TLS support via configmap and secrets

wallyqs:
Adds support to enable TLS for a cluster by mounting secrets into the NATS container with the certificates.  For this to work,  version 1.0.4 of the server has to be used as it is less aggressive when dns lookup fails, which can take some time on pod startup.  Example:

```
apiVersion: "nats.io/v1alpha1"
kind: "NatsCluster"
metadata:
  name: "nats"
spec:
  # Number of nodes in the cluster
  size: 3

  # Must use 1.0.4 in order to allow waiting for the A record
  # from nodes in the cluster to be ready
  version: "1.0.4"

  tls:
    # Certificates to secure the NATS client connections:
    serverSecret: "nats-clients-tls"

    # Certificates to secure the routes.
    routesSecret: "nats-routes-tls"
```


wallyqs:
Thanks @pires for taking a look, sounds good on changing to constants. Following up in a PR containing these fixes.

PR #29, opened 2018-03-09
Bump version of the API

wallyqs:
Bumps version of the operator and default NATS server to be used, plus some doc updates for the TLS with secrets feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,49 @@ NAME                   AGE
 example-nats-cluster   1s
 ```
 
+### TLS support
+
+By using a pair of opaque secrets (one for the clients and then another for the routes),
+it is possible to set TLS for the communication between the clients and also for the
+transport between the routes:
+
+```yaml
+apiVersion: "nats.io/v1beta1"
+kind: "NatsCluster"
+metadata:
+  name: "nats"
+spec:
+  # Number of nodes in the cluster
+  size: 3
+  version: "1.0.4"
+
+  tls:
+    # Certificates to secure the NATS client connections:
+    serverSecret: "nats-clients-tls"
+
+    # Certificates to secure the routes.
+    routesSecret: "nats-routes-tls"
+```
+
+In order for TLS to be properly established between the nodes, it is 
+necessary to create a wildcard certificate that matches the subdomain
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -8,7 +8,7 @@ NATS Operator manages NATS clusters atop [Kubernetes][k8s-home], automating thei
 
 ## Requirements
 
-- Kubernetes v1.7+
+- Kubernetes v1.8+
 - NATS Server v1.0.4+
```
pires: This is not longer a requirement, right?

example/deployment.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
       - name: nats-operator
-        image: nats-io/nats-operator:latest
+        image: wallyqs/nats-operator:0.2.0-v1beta1
```
pires: This doesn't look right to me.

ColinSullivan1: @pires , Great catch, thank you for noticing.  I asked @wallyqs to do this temporarily - Docker takes a *long* time to approve new images, and we'd like the community to be able to experiment with this in beta without having to build and use their own images for the operator. In the README, we should document that this image is used in the beta period, and only until the official docker image is available.  Note that for GA (v0.1.0) we'll be sure to use an official docker image.  We're certainly open to any other suggestions...

wallyqs:
Thanks @pires , ok so following a similar deprecation policy as in Kubernetes (https://kubernetes.io/docs/reference/deprecation-policy/) objects from v1alpha1 would then be deprecated (mainly the `tls` block which was inherited from the etcd-operator code), with an action required to uninstall previous `v1alpha1` version of the operator (which would remove the pods too) and then recreate new clusters using `v1alpha2` version instead. 

That is not very graceful but design around multiple CRD versions and objects migrations in k8s is still in flux so have to work around those constrains for now (https://github.com/kubernetes/kubernetes/issues/57316)

Also rather than jumping into `v1beta1`, we'd would bump into `v1alpha2` instead as that seems to fit better the current state, then bump into `v1beta1` once community has used it a bit more (PR updated to reflect this change).

PR #38, opened 2018-03-23
Add Authorization support

### LINKED ISSUE #1: Authentication support

### LINKED ISSUE #31: Enable customizing Authorization config
Add support to configure auth via the NatsCluster spec: https://github.com/nats-io/nats-operator/blob/master/pkg/conf/natsconf.go#L21

wallyqs:
This adds support for configuring users and permissions on subjects on a `NatsCluster`.  The users permissions have to be defined as a `Secret` in JSON which is then included as part of the configuration of the server:

```
kubectl create secret generic nats-clients-auth --from-file=clients-auth-permissions.json
```

Then as part of the `NatsCluster` definition, it is specified what is the secret from where the users permissions will be picked up:

```yaml
kind: "NatsCluster"
metadata:
  name: "example-nats-auth"
spec:
  size: 3
  version: "1.0.4"

  auth:
    # Definition in JSON of the users permissions
```

Fixes #1 
Fixes #31


pkg/conf/natsconf.go
```diff
@@ -39,30 +39,54 @@ type TLSConfig struct {
...
 var (
 	ErrInvalidConfig = errors.New("natsconf: cannot produce valid config")
 )
 
 func Marshal(conf *ServerConfig) ([]byte, error) {
-	js, err := json.MarshalIndent(conf, "", "  ")
+	buf := &bytes.Buffer{}
+	encoder := json.NewEncoder(buf)
+	encoder.SetEscapeHTML(false)
+	err := encoder.Encode(conf)
```
wallyqs: since `gnatsd` v1.1.0 is out, could wait for the Docker image to be available and remove all of this and just used JSON here (filed issue to remove: https://github.com/nats-io/nats-operator/issues/39)

PR #48, opened 2018-05-29
wip: PID namespace sharing based configuration reload

wallyqs:
This enables on-the-fly config reloading for the NATS Server by making use of a sidecar which exists its same PID namespace, so able to send HUP signal and trigger configuration reload.

This implementation is currently only supported in Kubernetes clusters that have enabled opt-in PID namespace (currently enabled by `--feature-gates=PodShareProcessNamespace=true`).  In order to enable configuration reload, this needs to be specified when creating the NATS cluster as part of the Pod policy as follows:

```yaml
kind: "NatsCluster"
metadata:
  name: "example-nats-auth"
spec:
  size: 5
  version: "1.1.0"

  pod:
    # Enable on-the-fly NATS Server config reload
    # Note: only supported in Kubernetes clusters with PID namespace sharing enabled.
    allowConfigReload: true

  auth:
    # Definition in JSON of the users permissions

    # How long to wait for authentication
    clientsAuthTimeout: 5
```

A secret named `clientsAuthSecret` can also be updated with the explicit configuration in JSON of the accounts in the server, also triggering reloading of the configuration, for example:

```js
{
  "users": [
    { "username": "user1", "password": "secret1" },
    { "username": "user2", "password": "secret2",
      "permissions": {
	"publish": ["hello.*"],
	"subscribe": ["hello.world"]
      }
    }
  ],
  "default_permissions": {
    "publish": ["SANDBOX.*"],
    "subscribe": ["PUBLIC.>"]
  }
}
```
```
kubectl create secret generic nats-clients-auth --from-file=clients-auth.json
```


pkg/reloader/reloader.go
```diff
@@ -0,0 +1,178 @@
...
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"strconv"
+	"syscall"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
```
ColinSullivan1: Do you need to add this to the gopkg manifest?

pkg/reloader/reloader.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	var (
+		err      error
+		proc     *os.Process
+		pid      int
+		attempts int
+	)
+	for {
+		pidfile, err := ioutil.ReadFile(r.PidFile)
```
ColinSullivan1: err is shadowed

pkg/cluster/reconcile.go
```diff
@@ -69,6 +69,10 @@ func (c *Cluster) reconcileSize(pods []*v1.Pod) error {
 		if err := c.removePod(pods[currentClusterSize-1].Name); err != nil {
 			return err
 		}
+		err := c.updateConfigMap()
```
ColinSullivan1: if err := ... ?

pkg/constants/constants.go
```diff
@@ -39,6 +39,19 @@ const (
...
+
+	// PidFileName is the pid file name.
+	PidFileName = "gnatsd.pid"
+
+	// PidFileMountPath is the absolute path to the directory where NATS
+	// will be leaving its pid file.
+	PidFileMountPath = "/var/run/nats"
+
+	// PidFilePath is the location of the pid file.
+	PidFilePath = PidFileMountPath + "/" + PidFileName
```
ColinSullivan1: Would it be simpler for users to just set the `PidFilePath` and you get the `PidFileMountPath` by parsing?  I don't see `PidFileName` used anyplace else.  wdyt?

pkg/spec/cluster.go
```diff
@@ -123,10 +123,11 @@ type PodPolicy struct {
...
 
 	// List of environment variables to set in the NATS container.
-	// This is used to configure NATS process. NATS cluster cannot be created, when
-	// bad environment variables are provided.
-	// This field cannot be updated.
 	NatsEnv []v1.EnvVar `json:"natsEnv,omitempty"`
+
+	// AllowConfigReload attaches a sidecar to each NATS Server
+	// that will signal the server whenever the configuration is updated.
+	AllowConfigReload bool `json:"allowConfigReload,omitempty"`
```
ColinSullivan1: `EnableConfigReload`?

pkg/util/kubernetes/pod.go
```diff
@@ -60,8 +59,22 @@ func natsPodContainer(clusterName, version string) v1.Container {
...
 		},
 	}
+}
 
-	return c
+// reloaderContainer returns a NATS server pod container spec.
+func natsPodReloaderContainer() v1.Container {
+	return v1.Container{
+		Name:            "pid",
+		Image:           "wallyqs/nats-server-config-reloader:0.2.1-v1alpha2",
```
ColinSullivan1: Should the image and the pull policy be able to be overridden?

pkg/reloader/reloader.go
wallyqs: ok already added this one

pkg/constants/constants.go
wallyqs: These aren't tunable for users so think ok just as constants for now

pkg/constants/constants.go
```diff
@@ -58,4 +71,10 @@ const (
 	RoutesCAFilePath     = RoutesCertsMountPath + "/ca.pem"
 	RoutesCertFilePath   = RoutesCertsMountPath + "/route.pem"
 	RoutesKeyFilePath    = RoutesCertsMountPath + "/route-key.pem"
+
+	// Default Docker Images
+	DefaultServerImage = "nats"
+	DefaultReloaderImage = "wallyqs/nats-server-config-reloader"
```
wallyqs: Will change this default image before releasing, but now made configurable

pkg/util/kubernetes/pod.go
wallyqs: These are now configurable:

```yaml
  pod:
    enableConfigReload: true
    reloaderImage: "connecteverything/nats-reloader"
    reloaderImageTag: "latest"
    reloaderImagePullPolicy: "Always"
```

pkg/reloader/reloader.go
wallyqs: should be fixed now

PR #58, opened 2018-07-10
Integration with ServiceAccounts using bound tokens

wallyqs:
Enables mapping ServiceAccounts from Kubernetes to be included as part of NATS Authentication by using the `TokenRequest API` (released as alpha in Kubernetes v1.10), in order to issue tokens with a different custom audience so that it cannot be used against the API server, but still coupled to the original ServiceAccount.

svc-account-bound-token

This is done by creating a new `NatsServiceRole` that defines the permissions for the account:

```yaml
---
apiVersion: nats.io/v1alpha2
kind: NatsCluster
metadata:
  name: example-nats
spec:
  size: 3
  version: "1.2.0"
  pod:
    enableConfigReload: true
  auth:
    enableServiceAccounts: true
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: nats-user
---
apiVersion: nats.io/v1alpha2
kind: NatsServiceRole
metadata:
  name: nats-user
  namespace: nats-io

  # Defines which cluster will be binding the account.
  labels:
    nats_cluster: example-nats
spec:
  permissions:
    publish: ["foo.*", "foo.bar.quux"]
    subscribe: ["foo.bar", "greetings", "hello.world"]
```

This will generate a new secret containing a token which can be used to auth against the NATS server, with the audience to only the service for the NATS cluster (`aud: [ "nats://example-nats.nats-io.svc" ]`).

A Pod from a user then can declare to mount the secret with the new token bound to the ServiceAccount to authenticate with NATS.

```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: nats-admin-user-pod-1
  labels:
    nats_cluster: example-nats
spec:
  volumes:
    - name: "token"
      projected:
        sources:
        - secret:
            name: "nats-admin-user-example-nats-bound-token"
            items:
              - key: token
                path: "token"
  containers:
    - name: nats-ops
      command: ["/bin/sh"]
      image: "wallyqs/nats-ops:latest"
      tty: true
      stdin: true
      stdinOnce: true
      volumeMounts:
      - name: "token"
        mountPath: "/var/run/secrets/nats.io"
        readOnly: true
``` 

```sh
nats-sub -s nats://nats-user:`cat /var/run/secrets/nats.io/token`@example-nats:4222 hello.world
Listening on [hello.world]
```

### TODO

- [x] Enable TokenRequest API based tests on Travis CI
- [x] Testing
- [x] Documentation
  + PR: https://github.com/nats-io/nats-operator/pull/59

### Demo

k8s-service-accounts-take-2


pkg/util/kubernetes/kubernetes.go
```diff
@@ -145,12 +147,129 @@ func addTLSConfig(sconfig *natsconf.ServerConfig, cs spec.ClusterSpec) {
...
+						UID:        tokenSecret.UID,
+					},
+				},
+			}
+			tr, err := kubecli.ServiceAccounts(ns).CreateToken(sa.Name, ar)
+			if err != nil {
+				return err
+			}
+
+			if err == nil {
```
ColinSullivan1: Not needed w/ return above.

pkg/util/kubernetes/kubernetes.go
```diff
@@ -145,12 +147,129 @@ func addTLSConfig(sconfig *natsconf.ServerConfig, cs spec.ClusterSpec) {
...
+	cs spec.ClusterSpec,
+	owner metav1.OwnerReference,
+) error {
 	if cs.Auth == nil {
 		return nil
 	}
-	if cs.Auth.ClientsAuthSecret != "" {
+
+	if cs.Auth.EnableServiceAccounts {
+		roleSelector := map[string]string{
```
ColinSullivan1: gometalinter complains that this function is too complex...   Maybe break it up a bit to improve readability?

ColinSullivan1:
On the new files, you'll want a 2018 copyright.

pkg/cluster/cluster.go
```diff
@@ -232,15 +237,62 @@ func (c *Cluster) run(stopC <-chan struct{}) {
...
-			// Just here get the secret version in case there is one.
 			if c.cluster.Spec.Auth != nil {
-				// Look for updates in the secret used for auth then
-				// trigger config reload in case there are new updates.
-				authSecret := c.cluster.Spec.Auth.ClientsAuthSecret
-				result, err := c.config.KubeCli.Secrets(c.cluster.Namespace).Get(authSecret, metav1.GetOptions{})
-				if err == nil && secretLastResourceVersion != result.ResourceVersion {
-					secretLastResourceVersion = result.ResourceVersion
-					c.updateConfigMap()
+				if c.cluster.Spec.Auth.ClientsAuthSecret != "" {
```
ColinSullivan1: Just a suggestion - you could move this into a checkCliientAuthUpdate function for readability...

pkg/cluster/cluster.go
```diff
@@ -232,15 +237,62 @@ func (c *Cluster) run(stopC <-chan struct{}) {
...
+					// Look for updates in the secret used for auth then
+					// trigger config reload in case there are new updates.
+					authSecret := c.cluster.Spec.Auth.ClientsAuthSecret
+					result, err := c.config.KubeCli.Secrets(c.cluster.Namespace).Get(authSecret, metav1.GetOptions{})
+					if err == nil && secretLastResourceVersion != result.ResourceVersion {
+						secretLastResourceVersion = result.ResourceVersion
+						c.updateConfigMap()
```
ColinSullivan1: Does this need an error check?

pkg/cluster/cluster.go
```diff
@@ -232,15 +237,62 @@ func (c *Cluster) run(stopC <-chan struct{}) {
...
+					for uid, _ := range cRoles {
+						if _, ok := aRoles[uid]; !ok {
+							shouldUpdate = true
+
+							// Stop tracking this role.
+							delete(cRoles, uid)
+						}
+					}
+					if shouldUpdate {
+						c.updateConfigMap()
```
ColinSullivan1: error check?

pkg/spec/cluster.go
```diff
@@ -142,8 +144,9 @@ type PodPolicy struct {
 // AuthConfig is the authorization configuration for
 // user permissions in the cluster.
 type AuthConfig struct {
-	ClientsAuthSecret  string `json:"clientsAuthSecret,omitempty"`
-	ClientsAuthTimeout int    `json:"clientsAuthTimeout,omitempty"`
+	EnableServiceAccounts bool   `json:"enableServiceAccounts,omitempty"`
```
ColinSullivan1: It may be good to add a few lines of doc about `EnableServiceAccounts` and `ClientsAuth` modes of auth and what they do.

Issue #61, opened 2018-08-01
Helm chart


richerlariviere:
First of all, thank you for your amazing work on Nats and particularly nats-operator.

What do you think of creating a [Helm](https://helm.sh/) chart for nats-operator? That would be nice I think because we could do something like this:

`helm install nats-operator`

I have some knowledge in creating helm charts but my global understanding of nats is not so great so I can help creating a base chart for a minimal setup. Edge cases and more specific uses of nats could be supported later. What do you think?

wallyqs:
A helm chart sounds good, feel free to send the PR for the basic setup :)

richerlariviere:
I have setup a repo here: https://github.com/richerlariviere/nats-cluster-chart. I tested the chart on my cluster and it worked well.

Currently the repo supports client authentication only via password as this is how I use NATS. I plan to add **token support** and the possibility to add other config files like for **cluster authentication**.

In fact, the repo is heavily inspired from the [NATS helm chart](https://github.com/helm/charts/tree/master/stable/nats). Eventually I'd like a configuration parity with the original chart. I also used the manifests in `/example` to build the templates.

When I'm done I'll let you know and we will be able to discuss how you want the chart to be included to the repo (I'm not sure where we could put it, maybe in a `contrib/` directory at the root?).

wallyqs:
Sorry for the wait, sounds good to include in the repo. I'm thinking the following directory structure, wdyt?

```sh
├── helm
│   └── nats-operator
│        ...
│       └── Chart.yaml
```
Thanks for the contribution!

richerlariviere:
That sounds good to me :)

wallyqs:
Fixed via #62

PR #62, opened 2018-08-07
Helm chart for nats-operator

richerlariviere:
Following our discussion (see issue #61 ), I decided to create a helm chart for nats-operator. I had some experience with Helm and my team is using NATS-operator (thanks a lot by the way). Despite my little understanding of NATS I think that this chart can cover the basic usage of NATS. So here is what I did so far:

- [x] Test cluster with authentication enabled
- [x] Test cluster with RBAC enabled
- [x] Test with 1 or more users
- [x] Test cluster with TLS 
- [x] Test cluster with configuration reload

 I'd like to have your feedback for other tests I could do.

I suggest that we don't upload this Helm on [KubeApps](https://hub.kubeapps.com/) until it's mature and well-tested.

**Thank you for your feedback!** 😀


helm/nats-operator/README.md
```diff
@@ -0,0 +1,139 @@
...
+NATS Operator manages [NATS](https://nats.io/) clusters atop [Kubernetes](http://kubernetes.io), automating their creation and administration. With NATS-operator you can benefits from the flexibility brought by the Kubernetes operator pattern. It means less juggling between manifests and a few handy features like automatic configuration reload.
+
+If you want to manage NATS entirely by yourself and have more control over your NATS cluster, you can always use the original [NATS](https://github.com/helm/charts/tree/master/stable/nats) Helm chart.
+
+
+## Prerequisites
+
+Basic
+- Kubernetes 1.4+ with Beta APIs enabled
+- PV provisioner support in the underlying infrastructure
```
wallyqs: sorry not following which Kubernetes feature is this one... we should just make the prerequisites same as in the nats operator readme (`Kubernetes v1.8+`, etc...)

helm/nats-operator/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+  #username: 
+  #password: 
+
+  # values.yaml
+  users:
+    - username: "my-user"
+      password: "T0pS3cr3t"
+    - username: "my-user-2"
+      password: "MyS3cr3tP4ssw0rd"
```
wallyqs: we could also declare `permissions:` here but could be done in a different PR

helm/nats-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+              fieldPath: metadata.namespace
+        - name: MY_POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+          {{- if .Values.livenessProbe.enabled }}
+          livenessProbe:
+            httpGet:
+              path: /
+              port: monitoring
```
wallyqs: Not sure that the NATS Operator has a monitoring port for liveness/readiness probe (may have missed it... if so we should include in the examples...)

helm/nats-operator/README.md
richerlariviere: I changed it si we only keep `Kubernetes v1.8+`. I don't know what I was thinking. I probably assumed wrongly that NATS and NATS-operator shared the same requirements in terms of Kubernetes configuration. I just pushed a [commit](c3a6cb5417f211920a671d4b718dd3c5ea3d9398) to fix that.

helm/nats-operator/README.md
richerlariviere: Yeah of course! I'll remind myself to create an issue after this one is merged.

helm/nats-operator/templates/deployment.yaml
wallyqs: Ok so we can use something for the readiness probe, adding to the examples: https://github.com/nats-io/nats-operator/pull/63

richerlariviere: I pushed a new commit changing the port to `8080`. I tested it in my side and I can poke `:8080/readyz` succesfully

helm/nats-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: MY_POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+          ports:
+          - name: readyz 
+            port: 8080
```
wallyqs: I think this one has to be `containerPort`? https://github.com/nats-io/nats-operator/blob/68079dc2093aba61711a8a19ae0989bd181b9cf4/example/deployment-rbac.yaml#L29-L35

```
        ports:
        - name: readyz
          containerPort: 8080
```

helm/nats-operator/README.md
richerlariviere: Issue created https://github.com/nats-io/nats-operator/issues/67

richerlariviere:
Added support for permissions. Fix #67

PR #63, opened 2018-08-08
Add readiness probe to operator deployment

wallyqs:
The operator process starts with an HTTP server that can serve as a readiness probe, so including as part of the deployment yaml examples:
https://github.com/nats-io/nats-operator/blob/b3aa986906b1396d1d7c37b883631042f898e5e6/cmd/operator/main.go#L148

### DIFF
--- a/example/deployment-rbac.yaml
+++ b/example/deployment-rbac.yaml
@@ -39,6 +39,12 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8080
+          initialDelaySeconds: 15
+          timeoutSeconds: 3
 ---
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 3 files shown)


example/deployment-rbac.yaml
```diff
@@ -39,6 +39,12 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8080
+          initialDelaySeconds: 15
+          timeoutSeconds: 3
```
richerlariviere: You could add 
```yaml
          ports:
          - name: readyz 
            port: 8080
```
That way the Dashboard can display the port like that:
screen shot 2018-08-08 at 1 51 04 pm

wallyqs: Weird, couldn't find it in my dashboard on minikube but a good idea so updated the example to use a named port.

Issue #49, opened 2018-06-03
service?


k1ng440:
How to create a service?

```
apiVersion: "nats.io/v1alpha2"
kind: "NatsCluster"
metadata:
  name: "cryptovalue-nats"
spec:
  size: 3
  version: "1.1.0"
```

wallyqs:
Just apply via `kubectl` and then you would get a `cryptovalue-nats` service to which pods can connect (e.g. `nats.Connect("nats://cryptovalue-nats:4222")`)
```
echo '
apiVersion: "nats.io/v1alpha2"
kind: "NatsCluster"
metadata:
  name: "cryptovalue-nats"
spec:
  size: 3
  version: "1.1.0"
' | kubectl apply -f -
```

MichaelSp:
How do I connect from outside the cluster (for development)?

wallyqs:
You need a configured .kube/config setup locally to authenticate with kubernetes first, check out the dev section on the readme
https://github.com/nats-io/nats-operator/blob/master/README.md#running-outside-the-cluster-for-debugging

MichaelSp:
Thanks for the answer @wallyqs 
My question was to unspecific. I wanted to connect to nats-cluster, not the nats-operator. And I didn't want to debug/develop nats-operator but rather a nats-enabled application. But I think I got a solution:

`kubectl port-forward ${CLUSTER_NAME}-1 4222:4222`

MichaelSp:
The hard thing was to find out the name of the pods. Maybe this one-liner in the README would also help others.

wallyqs:
@MichaelSp sounds good, feel free to make PR or I can update tomorrow. thanks for the feedback

PR #69, opened 2018-09-07
[doc] how to directly access to the cluster

MichaelSp:
as discussed in https://github.com/nats-io/nats-operator/issues/49


wallyqs:
Merged via #71

PR #79, opened 2018-10-23
Helm chart fix probes and resources

Will2817:
This PR addresses two problems:
1. The `port`, `livenessProbe` and `readinessProbe` were defined a level too deep under `env` instead of `container`.
2. The `resources` requests and limits of the `values.yaml` file were not being applied to the container

Describing the running pod should now produce the following:
image


wallyqs:
Thanks sharing the fix!

PR #91, opened 2018-11-11
controller: use informers to watch resources

pires:
**Description**

This PR changes the watch mechanism of `nats-operator` to use a shared index informer instead of the manual, error-prone process that exists today (which basically involves reading and decoding events from the Kubernetes API and manually keeping track of resource versions).

This PR is the first of a series of _rework_ changes that will be pushed to master branch altogether when the rework is done.

**Notes to reviewers**

- It's probably better to review this on a per-commit basis:
  - 6842bb1 just makes the e2e test suite use the `nats-operator` service account, as otherwise the suite won't run in an RBAC-enabled cluster.
  - 594d590 just pins dependencies to `v1.12.2` (instead of `master`).
  - 2d12833 looks big, but all it does is changing the code generation step to include generation of _listers_ and _informers_. The fact that `generate-groups.sh` makes assumptions about the location of the type definitions requires these to be moved from `pkg/spec` to `pkg/apis/nats/v1alpha2`. This changes import paths everywhere, but aligns us with the conventions out there today and makes code easier to follow. This commit also "unversions" all generated code.
  - 1906b81 replaces the aforementioned manual process of watching events with a shared index informer which does all the work for us. Since the pre-existing `handleClusterEvent` function is expecting an `Event` object, and in order to make this change as surgical as possible, we wrap the resource in an object of said type in `handleEvent`. This will be removed later on when we switch to using `util/workqueue`. This allows for removing a bunch of code and TODOs.
- This is to be merged into the `rework` branch, which will be rebased onto master after a series of PRs like this one.


.gitignore
```diff
@@ -7,3 +7,9 @@ vendor
 ## IntelliJ
 .idea
 *.iml
+
+## Generated code
+pkg/apis/nats/v1alpha2/zz_generated.deepcopy.go
```
wallyqs: I think ok to include the generated code no? Otherwise would have to generate for every build for example

pkg/client/cr.go
```diff
@@ -18,7 +18,7 @@ import (
 	"context"
 	"errors"
 
-	"github.com/nats-io/nats-operator/pkg/spec"
+	"github.com/nats-io/nats-operator/pkg/apis/nats/v1alpha2"
```
wallyqs: 👍

.gitignore
pires: Done.

pkg/controller/controller.go
```diff
@@ -71,81 +68,127 @@ type Config struct {
...
 func (c *Config) Validate() error {
 	return nil
 }
 
-func New(cfg Config) *Controller {
-	return &Controller{
+// NewNatsClusterController returns a new instance of a controller for NatsCluster resources.
+func NewNatsClusterController(cfg Config) *Controller {
+	// Create a shared informer factory for our API.
+	natsInformerFactory := natsinformers.NewSharedInformerFactory(cfg.OperatorCli, time.Second*30)
```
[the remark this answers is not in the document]
pires: Yes, I took down that feature request and we need to discuss it properly before implementing. Is it fine if this is merged as is for now? Do remember this is going to be merged into the rework branch, and not the master branch.

test/e2e/framework/framework.go
```diff
@@ -132,6 +131,7 @@ func (f *Framework) SetupNatsOperator() error {
 				},
 			},
 			RestartPolicy: v1.RestartPolicyNever,
```
wallyqs: For different PR but thinking might be better to change to `RestartPolicyAlways` in case they exit 1 for some reason, right now the operator does replace the pods only when they are missing.

wallyqs: For different PR but thinking might be better to change to `RestartPolicyAlways` in case they exit 1 for some reason, right now the operator does replace the pods only when they are missing.

pkg/controller/controller.go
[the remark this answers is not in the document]
pires: Maybe what we're looking for is `NewSharedInformerFactoryWithOptions(..., WithNamespace("foo"))`. I'll look into this in my next PR.

Issue #81, opened 2018-10-24
Support installing the CRD from k8s manifests instead of from the operator


seriousben:
Instead of creating the CRDs from the code, I think it would make sense to create it from k8s manifests while installing the operator itself. This could be done for the helm chart or for all installs.

This would allow:
* using the `crd-install` helm hook in helm charts
* getting rid of a rule in Role definitions when using rbac required to install a CRD:
```
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs: ["*"]
```

Implementation:
- support a command line flag to toggle crd install from the operator
- in the helm chart, disable crd installs from the operator
- in the helm chart, remove the rule for CRDs CRUD

wallyqs:
Agree we could creating the CRD make it part of the `deployment-rbac.yaml` example and remove some policy from the operator role.

PR #92, opened 2018-11-11
crds: revisit register and wait

pires:
This PR is the second of a series of rework changes that will be pushed to master branch altogether when the rework is done.

This PR improves the registration of the `NatsCluster` and the `NatsServiceRole` CRDs by introducing a "create or update" strategy. This allows the operator to update the CRDs without manual intervention, whenever required, and paves the way for the future introduction of CRD versioning and additional printer columns.

It also introduces a _wait_ step during CRDs registration, meaning the controller will only start processing resources once the `Established` condition has been appended to both CRDs.

Finally, this PR gets rid `time.Sleep(...)` calls in `test/operator` in favor of `WaitCRDs(...)`.

### DIFF
--- a/pkg/util/kubernetes/crd.go
+++ b/pkg/util/kubernetes/crd.go
@@ -16,23 +16,64 @@ package kubernetes
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
-	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"github.com/sirupsen/logrus"
+	extsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	extsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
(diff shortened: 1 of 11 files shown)


pkg/util/kubernetes/crd.go
```diff
@@ -16,23 +16,64 @@ package kubernetes
...
+)
+
+const (
+	// waitCRDReadyTimeout is the maximum period of time we wait for each CRD to be ready (i.e. established).
+	waitCRDReadyTimeout = 30 * time.Second
 )
 
 var (
-	ErrCRDAlreadyExists = errors.New("crd already exists")
+	// crds contains all the custom resource definitions that nats-operator registers upon starting.
```
wallyqs: Eventually we could look into having the operator not be responsible of creating these CRDs, would make it possible to narrow down some of the required policy for it, for now I added them to the manifests as well so that would be present already when deploying the operator.
https://github.com/nats-io/nats-operator/issues/81

pires: This doesn't make much sense to me as it limits what one can do with and to CRDs, like migration. Also, relying on the user seems to be exactly what we're avoiding by automating with an operator ;)

pires: The end-goal is for CRD multi-versioning, btw.

wallyqs: Right ok can be addressed when setting up for multi version CRD from k8s v1.11

Issue #90, opened 2018-11-09
Allow setting custom image for the NATS server


wallyqs:
Should allow making it possible to set a custom image for the server, for example:

```
apiVersion: "nats.io/v1alpha2"
kind: "NatsCluster"
metadata:
  name: "example-nats-1"
spec:
  size: 3
  # version "1.3.0"
  image: "nats:1.3.0"
```

onesolpark:
I think it would be great to add support for custom images..

I did made minimal changes to the codes to make it work in our environment 
(offline environment with no internet-access - Have to use a custom image stored in our private repo)

Will make a PR with the changes I made.

wallyqs:
Fixed via #93

PR #93, opened 2018-11-12
Added option to modify server image to custom image

onesolpark:
Regarding Issue #90 
Added serverImage option to natsCluster CRD to support customImage

An example CRD with custom image. 
```
kind: "NatsCluster"
metadata:
  name: "example-nats-1"
spec:
  size: 3
  version "1.3.0"
  serverImage: "customImage"
```

This will ultimately create a nats cluster using customImage:1.3.0 as the serverImage.  
DefaultServerImage in constants will be used(also set in CRD) if serverImage is not set.


wallyqs:
LGTM! Thanks for the contribution

PR #94, opened 2018-11-12
controller: use the work queue

pires:
This PR introduces support for using a [work queue](https://github.com/kubernetes/client-go/blob/master/util/workqueue/rate_limitting_queue.go) to distribute work among a number of worker threads (goroutines). This allows us to align with the [guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md#guidelines) for writing a controller by transitioning between the following two models:

**Current (`master`):**

nats-operator-master

**New (`rework`):**

nats-operator-rework

By transitioning between these models, we reduce the number of goroutines required to handle `NatsCluster` resources from `N` (one per cluster) to a constant value `C` (currently `2`), and also reduce the load on the Kubernetes API by reducing the number of requests issued against it.

A key part of this transition involves switching from a model in which we do a "unit of work" (e.g. create a single pod) every 5s to a model where a worker goroutine does _all the work_ required at the moment (e.g. create two pods to achieve a total size of five), becoming available for working on other `NatsCluster` resources until the "current" `NatsCluster` needs to be reconciled again. The fact that a `NatsCluster` needs to be reconciled is signaled by its `namespace/name` being added to the workqueue. In its turn, this happens whenever one of the following situations occurs:

1. The `NatsCluster` resource is _seen_ (e.g. created, or listed when the application starts);
1. The `NatsCluster` resource is modified.
1. A pod, service, secret or `NatsServiceRole` associated with the `NatsCluster` is seen, modified or deleted.
1. The full resync period of the shared informer factories (currently set to 24h) elapses. This may be disabled in the future. 

To allow for this change, we now wait for a pod to be ready after creating it, as well as for a pod to be effectively deleted after it is deleted. The order and naming strategy for pods is kept, even though `createPod` and `removePod` are being reworked to play nicely with listers - speaking of which, we now also use listers instead of directly querying the API to know the current state of a given resource. This also helps greatly reduce our request footprint.

For more info on _edge_ vs _level_ triggered reconciliation, please read [this article](https://hackernoon.com/level-triggering-and-reconciliation-in-kubernetes-1f17fe30333d).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,6 +10,14 @@
   version = "v0.32.0"
 
 [[projects]]
+  digest = "1:7e77c179d2f1df5c29550c9822dccda4ce31745e70464f16cacc713a1f0413ef"
+  name = "github.com/Azure/draft"
+  packages = ["pkg/kube/podutil"]
+  pruneopts = ""
+  revision = "5433afea1421810ae9d828631d8651de913b347a"
+  version = "v0.16.0"
+
+[[projects]]
   branch = "master"
   digest = "1:c0bec5f9b98d0bc872ff5e834fac186b807b656683bd29cb82fb207a1513fabb"
   name = "github.com/beorn7/perks"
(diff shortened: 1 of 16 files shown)


wallyqs:
Great that it is using more informers and listers 👍  will take a look further tomorrow.

pires:
@wallyqs please, re-review as I've added 6cbe8d8.

pires:
The test suites will be reworked after this PR is merged.

pires:
I've agreed with @wallyqs to merge this since it's not going to master and we discuss later.

PR #96, opened 2018-11-15
Create more strict rules for RBAC

metalmatze:
I don't think this operator needs all verbs on these Kubernetes objects. We should make sure to have less access.

This is possible missing something as I have reverse engineered this, rather than actually read through the code.

### DIFF
--- a/deploy/default-rbac.yaml
+++ b/deploy/default-rbac.yaml
@@ -38,12 +38,20 @@ rules:
 # Allow actions on basic Kubernetes objects
 - apiGroups: [""]
   resources:
-  - configmaps
-  - secrets
-  - pods
   - services
   - serviceaccounts
   - serviceaccounts/token
-  - endpoints
+  verbs: ["get"]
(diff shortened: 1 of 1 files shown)


deploy/default-rbac.yaml
```diff
@@ -38,12 +38,20 @@ rules:
...
+- apiGroups: [""]
+  resources:
   - events
-  verbs: ["*"]
+  verbs: ["create"]
+- apiGroups: [""]
+  resources:
+  - endpoints
+  - secrets
+  verbs: ["list", "get", "update"]
```
wallyqs: Thanks for the PR, agree on narrowing down the rbac policy examples. I think permissions to create secrets might be missing here

metalmatze: I'll update that and also the helm charts version tomorrow.

Smana:
Hi @metalmatze @wallyqs because yes the RBAC rules are clearly too open.

PR #105, opened 2018-12-06
ldm: introduce lame-duck mode support

pires:
This PR introduces support for recently introduced [`gnatsd` _lame-duck_ mode](https://github.com/nats-io/gnatsd/pull/780) (LDM).

### Changes

- LDM is achieved by running `/gnatsd -sl ldm=/var/run/nats/gnatsd.pid` from within the pod `nats` container.
- LDM duration is configurable via the new `.spec.lameDuckDuration` field.
- Enter LDM before removing or upgrading a pod, the operator will send a `SIGUSR2` signal to the `gnatsd` process running on said pod. 
- We're keeping the previous process of removing/upgrading pods (i.e. without gracefully shutting down active connections). This allows us to introduce support for LDM without breaking support for previous versions that do not support it.
-  In case LDM fails, either due to a networking problem or lack of support for LDM, the previous process of removing/upgrading pods is used.
- Quit the operator only when receiving the `SIGTERM` or `SIGINT` signals. Currently, the operators exits on any signal (including, notably, `SIGPIPE`).
- Fix Go context leaking in the end-to-end test suite.

### Notes to the reviewers

This PR requires a NATS image with the following (but not limited to) functionality:

- https://github.com/nats-io/gnatsd/pull/780
- https://github.com/nats-io/gnatsd/pull/799
- https://github.com/nats-io/gnatsd/pull/817
- https://github.com/nats-io/gnatsd/pull/838

After the `nats` container enters the `Terminated` state, upgrades must be performed by deleting and re-creating the pod. This happens because Kubernetes won't restart the container/pod if we simply change its image (as done in the current procedure). This effectively means that, in the future, all cluster upgrades will involve deleting and re-creating pods.

It should also be noted that no end-to-end test for the upgrade process is being added. This is due to the fact that we will need to wait for _two_ NATS releases before we can write a decent test. In fact, even when the next NATS version (say, `X`) is released, upgrading from `<= 1.3.0` to `X` will _not_ trigger LDM (since the source version does not support it). Hence, we can only test this scenario when NATS version `X+1` is out.

Finally, it is worth noting that scaling down/upgrading a cluster having `.spec.pod.enableConfigReload` set to `true` will [take a long time in older Kubernetes versions](https://github.com/kubernetes/kubernetes/issues/60518). According to my testing, this UX seems to have been fixed in more recent Kubernetes versions.
For affected Kubernetes versions, there's nothing the operator can do.


cmd/operator/main.go
```diff
@@ -95,7 +96,7 @@ func main() {
 	}
 
 	c := make(chan os.Signal, 1)
-	signal.Notify(c)
+	signal.Notify(c, syscall.SIGTERM, syscall.SIGINT)
```
wallyqs: 👍 we needed this (otherwise would exit even on resizing terminal)

deploy/default-rbac.yaml
```diff
@@ -41,6 +41,7 @@ rules:
   - configmaps
   - secrets
   - pods
+  - pods/exec
```
wallyqs: 👍

pkg/cluster/cluster.go
```diff
@@ -307,7 +330,9 @@ func (c *Cluster) createPod() (*v1.Pod, error) {
 
 	// Wait for the pod to be running and ready.
 	c.logger.Infof("waiting for pod %q to become ready", pod.Name)
-	if err := kubernetesutil.WaitUntilPodReady(c.config.KubeCli, pod); err != nil {
+	ctx, fn := context.WithTimeout(context.Background(), podReadinessTimeout)
+	defer fn()
+	if err := kubernetesutil.WaitUntilPodReady(ctx, c.config.KubeCli, pod); err != nil {
```
wallyqs: great, with context here will help with the timeout now and not block

pkg/util/kubernetes/exec.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return -1, "", "", err
+	}
+
+	// Capture stdout and stderr to dedicated buffers.
+	stdout := new(bytes.Buffer)
+	stderr := new(bytes.Buffer)
+	streamOptions := remotecommand.StreamOptions{
+		Stdin:  nil,
+		Stdout: stdout,
```
wallyqs: we could ioutil.Discard here instead of using the buffer

pires: I was already thinking of a future where stdout/stderr may become important, but that's not the case right now, so your suggestion makes a ton of sense.

pires: Done in 98435f0.

pkg/cluster/reconcile.go
```diff
@@ -45,7 +45,7 @@ func (c *Cluster) reconcileSize() error {
 		// Remove extra pods as required in order to meet the desired size.
 		// As we remove each pod, we must update the config secret so that routes are re-computed.
 		for idx := currentSize - 1; idx >= desiredSize; idx-- {
-			if err := c.removePod(pods[idx]); err != nil {
+			if err := c.tryGracefulPodDeletion(pods[idx]); err != nil {
```
wallyqs: maybe later we could look into being able to hook the operator so that on `kubectl delete natscluster` it calls deletion of the pods gracefully too

pires: @wallyqs can you please open a feature request issue describing the requirements?

PR #110, opened 2018-12-09
cluster/client: advertise pod hostname

### LINKED ISSUE #32: Add Cluster/Client Advertise to cluster config
Currently when TLS is setup for clients and routes, the NATS clients receive routes as IPs via auto discovery so hostname verification would fail when reconnecting to one of these IPs.  Setting instead the pods to advertise the A record that they get in the cluster would allow clients to be able to failover to an available node right away.

https://github.com/nats-io/gnatsd/blob/163ba3f6a7521300a4f578eb90b9ff5d40d658c4/main.go#L24

https://github.com/nats-io/gnatsd/blob/163ba3f6a7521300a4f578eb90b9ff5d40d658c4/main.go#L51

pires:
Fixes #32

### DIFF
--- a/pkg/util/kubernetes/kubernetes.go
+++ b/pkg/util/kubernetes/kubernetes.go
@@ -548,7 +548,7 @@ func addOwnerRefToObject(o metav1.Object, r metav1.OwnerReference) {
 }
 
 // NewNatsPodSpec returns a NATS peer pod specification, based on the cluster specification.
-func NewNatsPodSpec(name, clusterName string, cs v1alpha2.ClusterSpec, owner metav1.OwnerReference) *v1.Pod {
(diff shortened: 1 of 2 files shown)


pkg/util/kubernetes/kubernetes.go
```diff
@@ -598,13 +598,21 @@ func NewNatsPodSpec(name, clusterName string, cs v1alpha2.ClusterSpec, owner met
...
 	cmd := []string{
 		constants.NatsBinaryPath,
 		"-c",
 		constants.ConfigFilePath,
 		"-P",
 		constants.PidFilePath,
+		"--cluster_advertise",
+		advertiseHost,
+		"--client_advertise",
+		advertiseHost,
```
wallyqs: cluster advertise is correct but client advertise won't be right in some cases (for example when setting the client ip to be the one from the kubelet node, though will work in some cases when the traffic is all internal)

pires:
IIUC this was merged as part of a different PR. Closing!

PR #112, opened 2018-12-20
Add metrics port to mgmt service

arminbuerkle:
While adding prometheus annotations for metric scraping is good enough for the usual prometheus deployment, the prometheus operator requires a service target for its service monitor. 

https://github.com/coreos/kube-prometheus/pull/16#issuecomment-305933103

---

This PR exposes the `metrics` port on the nats mgmt svc, which can then be scraped by a `coreos ServiceMonitor`:
```yaml
---
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    app: nats
  name: nats
spec:
  endpoints:
    - interval: 30s
      port: metrics
  jobLabel: nats
  selector:
    matchLabels:
      app: nats
```

### DIFF
--- a/pkg/util/kubernetes/kubernetes.go
+++ b/pkg/util/kubernetes/kubernetes.go
@@ -128,6 +128,12 @@ func CreateMgmtService(kubecli corev1client.CoreV1Interface, clusterName, cluste
 			TargetPort: intstr.FromInt(constants.MonitoringPort),
 			Protocol:   v1.ProtocolTCP,
 		},
+		{
+			Name:       "metrics",
+			Port:       constants.MetricsPort,
+			TargetPort: intstr.FromInt(constants.MetricsPort),
+			Protocol:   v1.ProtocolTCP,
+		},
 	}
 	selectors := LabelsForCluster(clusterName)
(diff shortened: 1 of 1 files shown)


pkg/util/kubernetes/kubernetes.go
```diff
@@ -128,6 +128,12 @@ func CreateMgmtService(kubecli corev1client.CoreV1Interface, clusterName, cluste
 			TargetPort: intstr.FromInt(constants.MonitoringPort),
 			Protocol:   v1.ProtocolTCP,
 		},
+		{
+			Name:       "metrics",
+			Port:       constants.MetricsPort,
```
arminbuerkle: It'll add a metrics port regardless whether metrics are enabled or not on the crd.

While this isn't optimal, the current code doesn't give the option to update the service after it is created. Meaning if someone enables metrics after deploying a nats cluster, it won't be updated.

---
Additionally AFAIK k8s doesn't add the pods to the service/metrics endpoint which do not expose their metrics pod.

Therefore if metrics are disabled, no endpoints will be added, which is the correct behaviour.

PR #123, opened 2019-02-06
Add support for host port and no advertise

wallyqs:
Adds support for host port to be able to bind a nats server to an external ip, plus support for disabling advertising the internal ips to clients that connect externally.

```yaml
kind: "NatsCluster"
metadata:
  name: "example-nats-1"
spec:
  size: 3
  version: "1.4.0"
  noAdvertise: true
  pod:
    enableClientsHostPort: true
```

### DIFF
--- a/pkg/util/kubernetes/pod.go
+++ b/pkg/util/kubernetes/pod.go
@@ -36,8 +36,8 @@ import (
 )
 
 // natsPodContainer returns a NATS server pod container spec.
-func natsPodContainer(clusterName, version string, serverImage string) v1.Container {
-	return v1.Container{
+func natsPodContainer(clusterName, version string, serverImage string, enableClientsHostPort bool) v1.Container {
+	container := v1.Container{
 		Env: []v1.EnvVar{
 			{
 				Name:  "SVC",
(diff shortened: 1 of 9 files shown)


pkg/util/kubernetes/pod.go
```diff
@@ -50,24 +50,33 @@ func natsPodContainer(clusterName, version string, serverImage string) v1.Contai
...
-				Protocol:      v1.ProtocolTCP,
-			},
-			{
-				Name:          "monitoring",
-				ContainerPort: int32(constants.MonitoringPort),
-				Protocol:      v1.ProtocolTCP,
-			},
+	}
+
+	ports := []v1.ContainerPort{
```
ColinSullivan1: This change is for the case no monitoring or routes are setup, right?

ColinSullivan1:
LGTM.

pkg/util/kubernetes/pod.go
wallyqs: yes those are the defaults for cluster and monitoring ports, pnly change is for the client port to be able to bind the host port below that can be used to connect to it using the host network: `port.HostPort = int32(constants.ClientPort)`

PR #127, opened 2019-02-21
Enable Client Advertise for External IPs

### LINKED ISSUE #120: Exposing the NATS server with an External IP
In my use case, subscribers can come from outside the K8s cluster, and thus I need an external IP for the messaging port. What would be the right design to accomplish this in the chart as a configurable parameter?

wallyqs:
Adds support for the NATS server container to be able to advertise the external IP address of the node where it is running on.   In order to do this, it relies on an extra initializer container named 'bootconfig' that queries the Kubernetes API with the name of the node injected as an environment variable, then writes it as a file which is then mounted as a volume by the pod and the loaded via the `include` directive from the NATS server configuration file.

Also a `template` parameter is added in order to be able to parameterize a `NatsCluster` a la `Deployment`s to be able to set a service account, which is required in order to be able to peek at the external IP metadata from the server.

A full example of this advanced configuration is below:

```yaml
kind: "NatsCluster"
metadata:
  name: "nats-dev"
spec:
  size: 3
  version: "1.4.0"

  pod:
    enableClientsHostPort: true
    advertiseExternalIP: true
    bootconfigImage: "wallyqs/nats-boot-config"
    bootconfigImageTag: "0.5.0"

  template:
    spec:
      serviceAccountName: "nats-server"
```

Fixes #120

### DIFF
--- a/cmd/bootconfig/main.go
+++ b/cmd/bootconfig/main.go
@@ -1,0 +1,64 @@
+// Copyright 2018 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+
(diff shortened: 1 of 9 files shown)


wallyqs:
This mechanism might be a bit sophisticated but there does not seem to be another mechanism for the container to be able to reflect what is the external ip of the node.  There is host ip available but this is an internal ip most of the time... :/

cmd/bootconfig/main.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2018 The NATS Authors
```
ColinSullivan1: I wonder if at some point (not this PR) a special build of the NATS server into a process that includes the sidecars/helpers might be cleaner.

wallyqs: I like this idea a lot...

PR #130, opened 2019-02-22
Support for extra routes

wallyqs:
Allows being to map routes that are not part of the cluster, which can be useful for blue green deployments and draining scenarios.  Also includes support for setting cluster advertise from #110.

```yaml
kind: "NatsCluster"
metadata:
  name: "nats-dev-v2"
spec:
  size: 3
  version: "1.4.0"

  extraRoutes:
    - route: "nats://example.com:6222"
    - cluster: "nats-dev-v1"
```


pkg/constants/constants.go
```diff
@@ -30,6 +30,9 @@ const (
 	// MetricsPort is the port for the prometheus metrics endpoint.
 	MetricsPort = 7777
 
+	// ConnectRetries is the number of retries for an implicit route.
+	ConnectRetries = 10
```
ColinSullivan1: Does this reconcile with the default in the NATS server?

test/e2e/framework/natscluster.go
```diff
@@ -194,7 +206,10 @@ func (f *Framework) PatchCluster(natsCluster *natsv1alpha2.NatsCluster) (*natsv1
 // VersionForPod returns the version of NATS reported by the specified pod.
 func (f *Framework) VersionForPod(pod v1.Pod) (string, error) {
 	// Check the "/varz" endpoint for the number of established routes.
-	r, err := http.Get(fmt.Sprintf("http://%s:%d/varz", pod.Status.PodIP, constants.MonitoringPort))
+	client := &http.Client{
```
ColinSullivan1: Do we need to worry about https here?

pkg/constants/constants.go
wallyqs: This is higher than the default from the NATS server, which is only a couple of times.  It takes sometime in Kubernetes for DNS service to be ready so sometimes routes would have been dropped

test/e2e/framework/natscluster.go
wallyqs: Yes, this test is not using https for this check but if https monitoring is enabled then this has to be https.

PR #132, opened 2019-02-25
Add verify and map auth support

wallyqs:
Enables verify and map support that is available in edge from the [nats server](https://github.com/nats-io/gnatsd#tls-authorization)

```yaml
kind: "NatsCluster"
metadata:
  name: "nats-cluster"
spec:
  size: 3
  serverImage: "wallyqs/nats-server"
  version: "edge-2.0.0-RC5"

  tls:
    serverSecret: "nats-clients-tls"

  auth:
    tlsVerifyAndMap: true
```

### DIFF
--- a/pkg/apis/nats/v1alpha2/cluster.go
+++ b/pkg/apis/nats/v1alpha2/cluster.go
@@ -239,8 +239,8 @@ type AuthConfig struct {
 	// allow to clients to send their auth credentials.
 	ClientsAuthTimeout int `json:"clientsAuthTimeout,omitempty"`
 
-	// EnableTLSAuth toggles the verify and map.
-	EnableTLSAuth bool `json:"enableTLSAuth,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/apis/nats/v1alpha2/cluster.go
```diff
@@ -238,6 +238,9 @@ type AuthConfig struct {
 	// ClientsAuthTimeout is the time in seconds that the NATS server will
 	// allow to clients to send their auth credentials.
 	ClientsAuthTimeout int `json:"clientsAuthTimeout,omitempty"`
+
+	// EnableTLSAuth toggles the verify and map.
+	EnableTLSAuth bool `json:"enableTLSAuth,omitempty"`
```
ColinSullivan1: imo we should keep this inline with NATS server config nomenclature.  Maybe TLSVerifyAndMap?  Wdyt?

wallyqs: sounds good to me! I will update to use that instead

PR #134, opened 2019-03-01
Add field to include extra config for the server

wallyqs:
Add field to be able to enable debug and trace, as well as to customize max conns/payload and slow consumer deadline.

```yaml
kind: "NatsCluster"
metadata:
  name: "example-nats"
spec:
  size: 3
  version: "1.4.0"

  # Extra configuration for the server.
  natsConfig:
    debug: true
    trace: true

    # Duration within quotes
    write_deadline: "5s" 

    # In bytes, in this case 5MB
    max_payload: 5242880

    max_connections: 1024
    max_subscriptions: 1024
    max_pending: 65536 # In bytes
    disable_logtime: true
    max_control_line: 2048
```


pkg/apis/nats/v1alpha2/cluster.go
```diff
@@ -126,6 +130,15 @@ type ClusterSpec struct {
 	ExtraRoutes []*ExtraRoute `json:"extraRoutes,omitempty"`
 }
 
+// ServerConfig is extra configuration for the NATS server.
+type ServerConfig struct {
+	Debug          bool   `json:"debug,omitempty"`
+	Trace          bool   `json:"trace,omitempty"`
+	WriteDeadline  string `json:"write_deadline,omitempty"`
+	MaxConnections int    `json:"max_connections,omitempty"`
```
ColinSullivan1: While at it, maybe add MaxSubscriptions, MaxControlLine, LogTime?  Wdyt?

wallyqs: sgtm, will too

wallyqs: Added more extra fields for the limits, thanks for taking a look.

PR #145, opened 2019-03-12
Helm: restrict RBAC namespaced deployments

Smana:


helm/nats-operator/templates/rbac.yaml
```diff
@@ -0,0 +1,76 @@
+{{- if .Values.rbacEnabled }}
+apiVersion: rbac.authorization.k8s.io/v1
```
wallyqs: Thanks so much for the help with the Helm templates! Maybe we could update to the revised rbac policy that has less permissions now? https://github.com/nats-io/nats-operator/blob/master/deploy/00-prereqs.yaml#L37-L97

Smana: Changing that right now

Smana: done

PR #148, opened 2019-03-14
[Helm] rbac add missing perms

Smana:


helm/nats-operator/templates/rbac.yaml
```diff
@@ -72,14 +72,13 @@ rules:
   resources:
   - namespaces
   - serviceaccounts
-  verbs:
-  - list
+  verbs: ["list", "get", "watch"]
```
wallyqs: thanks, looks like I'm missing them as well here: https://github.com/nats-io/nats-operator/blob/master/deploy/00-prereqs.yaml#L58-L74
I'll add them after merging

PR #150, opened 2019-03-22
Remove garbage collection in the background

wallyqs:
Since we can rely on cascade ownership delete instead to delete the resources.


pires:
Why not remove the package entirely? Also, please make sure to update vendor, just in case.

PR #151, opened 2019-03-22
Change prometheus-nats-exporter to version 0.2.0 and add channelz and…

pslijkhuis:
… serverz to the arguments of the metrics container

Make nats exporter 0.2.0 the default exporter and add the arguments -serverz and -channelz

### DIFF
--- a/pkg/util/kubernetes/pod.go
+++ b/pkg/util/kubernetes/pod.go
@@ -114,6 +114,8 @@ func natsPodMetricsContainer(image, tag, pullPolicy string) v1.Container {
 			"-routez",
 			"-subz",
 			"-varz",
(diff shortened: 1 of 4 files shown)


pkg/util/kubernetes/pod.go
```diff
@@ -114,6 +114,8 @@ func natsPodMetricsContainer(image, tag, pullPolicy string) v1.Container {
 			"-routez",
 			"-subz",
 			"-varz",
+			"-channelz",
+			"-serverz",
```
wallyqs: btw these two flags will only work for the monitoring endpoint of the streaming server which this operator does not support yet

PR #155, opened 2019-04-01
Reverse order of CRD init operations

schancel:
This commit reverses the order of CRD init operations. The reason for
doing CRD init operations in this new order, is so that k8s admins
can manually create the expected CRDs. By creating the CRDs by hand,
create CRD permission does not need to be granted to the nats-operator SA.
Only read permission is required.

### DIFF
--- a/pkg/util/kubernetes/crd.go
+++ b/pkg/util/kubernetes/crd.go
@@ -128,36 +128,25 @@ func WaitCRDs(extsClient extsclientset.Interface) error {
 
 // createOrUpdateCRD creates or updates the specified custom resource definition according to the provided specification.
 func createOrUpdateCRD(crd *extsv1beta1.CustomResourceDefinition, extsClient extsclientset.Interface) error {
-	// Attempt to register the CRD.
-	_, err := extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd)
-	if err == nil {
-		// Registration was successful.
+	// At this point the CRD may already exist from manual creation. Attempt to get the CRD
+	d, err := extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(crd.Name, metav1.GetOptions{})
+
+	// CRD already exists, but it's what we expect.
+	if err == nil && reflect.DeepEqual(d.Spec, crd.Spec) {
 		return nil
 	}
-	if !IsKubernetesResourceAlreadyExistError(err) {
-		// The crd doesn't exist yet but we've got an unexpected error while registering it.
+
+	// CRD already exists, and is different than what is expected.
+	if err == nil {
+		// Attempt to update the CRD by setting its spec to the expected value.
+		d.Spec = crd.Spec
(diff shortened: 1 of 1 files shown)


pkg/util/kubernetes/crd.go
```diff
@@ -128,35 +128,28 @@ func WaitCRDs(extsClient extsclientset.Interface) error {
...
 	}
-	if !IsKubernetesResourceAlreadyExistError(err) {
-		// The crd doesn't exist yet but we've got an unexpected error while registering it.
-		return err
+
+	// CRD already exists, and is different than what is expected.
+	if err == nil {
+		// Attempt to update the CRD by setting its spec to the expected value.
+		d.Spec = crd.Spec
+		return extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Update(d)
```
wallyqs:
```suggestion
		_, err = extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Update(d)
		return err
```

pkg/util/kubernetes/crd.go
```diff
@@ -128,35 +128,28 @@ func WaitCRDs(extsClient extsclientset.Interface) error {
...
 	}
 
-	// At this point the CRD already exists but its spec may differ.
-	// This can happen, for instance, if the user has tampered with the CRD.
-	// Hence, we do our best to bring in line with our expectations.
-	// Fetch the latest version of the CRD.
-	d, err := extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(crd.Name, metav1.GetOptions{})
+	// No CRD existed, attempt to register the CRD.
+	_, err := extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd)
```
wallyqs:
```suggestion
	_, err = extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Create(crd)
```

wallyqs:
Thanks @schancel for the PR! Left a couple of suggestions that I think will fix the error in the build.

schancel:
@wallyqs Should be fixed.  Any chance you're willing to switch to `go mod`?  It would make it easier to build.  I can't actually build this right now, without changing my go setup.

pkg/util/kubernetes/crd.go
```diff
@@ -128,35 +128,29 @@ func WaitCRDs(extsClient extsclientset.Interface) error {
...
-		return err
-	}
-	if reflect.DeepEqual(d.Spec, crd.Spec) {
-		// The specs match, so there's nothing to do.
-		return nil
-	}
-	// Attempt to update the CRD by setting its spec to the expected value.
-	d.Spec = crd.Spec
-	if _, err := extsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Update(d); err != nil {
+		// Registration failed
```
schancel: Actually this is silly...

wallyqs:
Sorry about the inconvenience, will switch to use go modules after this.

Issue #156, opened 2019-04-01
Add support for `clientAuthFile`


schancel:
Add a `auth` config item `clientAuthFile` which tells nats operator to adjust the location of the nats auth configuration file.  The purpose of this would be for cases where secrets are coming from an `initContainer` defined in the `NatsCluster` definition.

schancel:
You can assign this to me, I'm going to submit a pull request.

wallyqs:
Fixed via #159

PR #157, opened 2019-04-02
Allow Pod Template to provide initContainers, Volumes, and Containers

schancel:
Previously, `NewNatsPodSpec` would initialize it the spec from a deep
copy of the PodTemplate, if it was provided by the user.  However, it then
overwrites initContainers, Volumes, and Containers with item lists that
are build at runtime.

This commit changes the behavior to append to the existing lists, enabling
InitContainers, Volumes, and Containers to be specified in the pod template.

Additionally, this adds a `VolumeMounts` item to the PodPolicy so that
volumes provided in the template, can be mounted into the nats container.

Step #1 of addressing https://github.com/nats-io/nats-operator/issues/156


wallyqs:
Merged including a couple of fixes at https://github.com/nats-io/nats-operator/commit/bbff7978648d4411c157c16bded3df51a9a017cf

PR #158, opened 2019-04-02
Adopt Go modules

wallyqs:
Setup Go modules and vendoring for the repo to reproduce the build


wallyqs:
Almost done but need to look why now with the updated deps after running go modules this is failing... 😶 

```
=== RUN   TestConfigReloadOnNatsServiceRoleUpdates
--- FAIL: TestConfigReloadOnNatsServiceRoleUpdates (11.67s)
    config_reload_test.go:338: expected subscription to "hello.world" to have failed
```

wallyqs:
Going to wait for Kubernetes v1.15 until this plan has been adopted: https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-03-19-go-modules.md#motivation
Otherwise will end up with only v0.0.0 pseudo version since upstream not using Go modules to avoid changing the imports across repos which is a bit tricky

wallyqs:
Closing for now, may reopen soon

PR #159, opened 2019-04-02
Add support for reading Auth data from a file on disk

schancel:
This commit adds support from reading the authorization section of the
gnatsd config from a JSON file located within the nats-operator container.
This allows secrets the data to be placed on disk using sidecars like
`consul-template`.

Depends on https://github.com/nats-io/nats-operator/pull/171

### DIFF
--- a/example/example-nats-cluster-authfile.yaml
+++ b/example/example-nats-cluster-authfile.yaml
@@ -1,0 +1,35 @@
+# This is an example NatsCluster manifest which uses a 3rd party initContainer
+# to fetch the authorization credentials from outside kubernetes.
+#
+# An example of this could be consul-template getting user/passwords from vault.
+apiVersion: "nats.io/v1alpha2"
+kind: "NatsCluster"
+metadata:
+  name: my-nats-cluster
+  namespace: nats-stuff
+spec:
+  size: 5
+  version: "1.3.0"
+  service:
+    maxPayload: 20971520
+  pod:
+    volumeMounts:
+      - name: authconfig
+        mountPath: /etc/nats-config/authconfig
+  template:
+    serviceAccountName: my-secrets-getting-sa
+    initContainers:
+      - name: secret-getter
+        image: "busybox"
+        command: ["sh", "-c", "echo 'users = [ { user: 'foo', pass: 'bar' } ]' > /etc/nats-config/authconfig/auth.json"]
+        volumeMounts:
+          - name: authconfig
+            mountPath: /etc/nats-config/authconfig
+    volumes:
+      - name: authconfig
+        emptyDir: {}
+  auth:
+    # Needs to be under /etc/nats-config where nats looks
+    # for its config file, or it won't be able to be included
+    # by /etc/nats-config/gnatsd.conf
+    clientAuthFile: "authconfig/auth.json"
--- a/pkg/reloader/reloader.go
+++ b/pkg/reloader/reloader.go
@@ -9,7 +9,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
-	"path"
+	"path/filepath"
 	"strconv"
 	"syscall"
 	"time"
@@ -20,7 +20,7 @@ import (
 // Config represents the configuration of the reloader.
 type Config struct {
 	PidFile       string
-	ConfigFile    string
+	ConfigFiles   []string
 	MaxRetries    int
 	RetryWaitSecs int
 }
@@ -39,10 +39,6 @@ type Reloader struct {
 
 	// quit shutsdown the reloader.
 	quit func()
-
-	// lastAppliedVersion is the last config update
-	// done by the proces..
-	lastAppliedVersion []byte
 }
 
 // Run starts the main loop.
@@ -94,41 +90,80 @@ func (r *Reloader) Run(ctx context.Context) error {
 	// Follow configuration updates in the directory where
 	// the config file is located and trigger reload when
 	// it is either recreated or written into.
-	if err := configWatcher.Add(path.Dir(r.ConfigFile)); err != nil {
-		return err
+	for i := range r.ConfigFiles {
+		// Ensure our paths are canonical
+		r.ConfigFiles[i], _ = filepath.Abs(r.ConfigFiles[i])
+		// Use directory here because k8s remounts the entire folder
+		// the config file lives in. So, watch the folder so we properly receive events.
+		if err := configWatcher.Add(filepath.Dir(r.ConfigFiles[i])); err != nil {
+			return err
+		}
 	}
 
 	attempts = 0
+	// lastConfigAppliedCache is the last config update
+	// applied by us
+	lastConfigAppliedCache := make(map[string][]byte)
+
+	// Preload config hashes, so we know their digests
+	// up front and avoid potentially reloading when unnecessary.
+	for _, configFile := range r.ConfigFiles {
+		h := sha256.New()
+		f, err := os.Open(configFile)
+		if err != nil {
+			return err
+		}
+		if _, err := io.Copy(h, f); err != nil {
+			return err
+		}
+		digest := h.Sum(nil)
+		lastConfigAppliedCache[configFile] = digest
+	}
+
+WaitForEvent:
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
 		case event := <-configWatcher.Events:
 			log.Printf("Event: %+v \n", event)
-			// FIXME: This captures all events in the same folder, should
-			// narrow down to updates to the config file involved only.
-			if event.Op != fsnotify.Write && event.Op != fsnotify.Create {
+			touchedInfo, err := os.Stat(event.Name)
+			if err != nil {
 				continue
 			}
 
-			h := sha256.New()
-			f, err := os.Open(r.ConfigFile)
-			if err != nil {
-				log.Printf("Error: %s\n", err)
-				continue
-			}
-			if _, err := io.Copy(h, f); err != nil {
-				log.Printf("Error: %s\n", err)
-				continue
-			}
-			digest := h.Sum(nil)
-			if r.lastAppliedVersion != nil {
-				if bytes.Equal(r.lastAppliedVersion, digest) {
-					// Skip since no meaningful change
+			for _, configFile := range r.ConfigFiles {
+				configInfo, err := os.Stat(configFile)
+				if err != nil {
+					log.Printf("Error: %s\n", err)
+					continue WaitForEvent
+				}
+				if !os.SameFile(touchedInfo, configInfo) {
 					continue
 				}
+
+				h := sha256.New()
+				f, err := os.Open(configFile)
+				if err != nil {
+					log.Printf("Error: %s\n", err)
+					continue WaitForEvent
+				}
+				if _, err := io.Copy(h, f); err != nil {
+					log.Printf("Error: %s\n", err)
+					continue WaitForEvent
+				}
+				digest := h.Sum(nil)
+				lastConfigHash, ok := lastConfigAppliedCache[configFile]
+				if ok && bytes.Equal(lastConfigHash, digest) {
+					// No meaningful change or this is the first time we've checked
+					continue WaitForEvent
+				}
+				lastConfigAppliedCache[configFile] = digest
+
+				// We only get an event for one file at a time, we can stop checking
+				// config files here and continue with our business.
+				break
 			}
-			r.lastAppliedVersion = digest
 
 		case err := <-configWatcher.Errors:
 			log.Printf("Error: %s\n", err)
--- a/pkg/util/kubernetes/pod.go
+++ b/pkg/util/kubernetes/pod.go
@@ -20,7 +20,7 @@ import (
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -80,8 +80,8 @@ func natsPodContainer(clusterName, version string, serverImage string, enableCli
 }
 
 // natsPodReloaderContainer returns a NATS server pod container spec for configuration reloader.
-func natsPodReloaderContainer(image, tag, pullPolicy string) v1.Container {
-	return v1.Container{
+func natsPodReloaderContainer(image, tag, pullPolicy, authFilePath string) v1.Container {
+	container := v1.Container{
 		Name:            "reloader",
 		Image:           fmt.Sprintf("%s:%s", image, tag),
 		ImagePullPolicy: v1.PullPolicy(pullPolicy),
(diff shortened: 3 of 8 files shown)


schancel:
Untested, and needs a test.  What do you recommend here @wallyqs ?

pkg/cluster/cluster.go
```diff
@@ -188,6 +191,18 @@ func (c *Cluster) checkClientAuthUpdate() error {
 			c.cluster.SetClientAuthSecretResourceVersion(result.ResourceVersion)
 			return c.updateConfigSecret()
 		}
+	} else if c.cluster.Spec.Auth.ClientsAuthFile != "" {
+		clientAuthBytes, err := ioutil.ReadFile(c.cluster.Spec.Auth.ClientsAuthFile)
```
schancel: The nasty thing about this, it it requires the auth file to be attached to the nats-operator.  So you need to be able to modify it's deployment to change a NatsCluster.  Seems like bad separation.

Originally, I wanted to make this adjust the container cmd to just be `gnatsd -c {{ authFile }}`` However, then all the other configuration options that nats-operator handles gets overridden.  gnatsd doesn't seem to support merging multiple config files together like consul, etc.

schancel:
Implements https://github.com/nats-io/nats-operator/issues/156 ?

pkg/cluster/cluster.go
wallyqs: I think maybe what we can do is to instead use the `Include` option to point to the path of the secret that was mounted via the volume mounts merged from the prev PR:　https://github.com/nats-io/nats-operator/blob/master/pkg/conf/natsconf.go#L27
That would generalize this as well into being able to opt-in override other parts of the config as well if needed (new options in the config file would not override others already done by the operator)

wallyqs: Here `Include` is being used to include a generated config file from another container, in this case only containing the advertise external IP address: 
https://github.com/nats-io/nats-operator/blob/bbff7978648d4411c157c16bded3df51a9a017cf/pkg/util/kubernetes/kubernetes.go#L413-L432

This would also mean that if wanting to use a custom volume mount to include a config and also have the external ip address, the config file in the custom volume mount need to also have an `include` to the generated file with the ip address (which I think is ok).

wallyqs: Then since it is a mounted file, then the reloader sidecar could track the changes to the file instead of the operator by adding the custom volume mount secret config file here: https://github.com/nats-io/nats-operator/blob/master/pkg/reloader/reloader.go#L94-L99

schancel: Okay, I misunderstood what you're saying, but I think this is better:

1. Add an include section into the `AuthConfig` to allow users to specify a file to include. 
2. Then using the template, and VolumeMounts stuff from the previous patch, we are able to add an:
    a.  a shared empty volume
    a. initContainer that drops a auth config json file into above volume
    b. use the VolumeMounts option to mount that volume into the nats container.
    c.  It should include the auth config file from there.

So this would result in a NatsCluster CRD like this:

```yaml
kind: "NatsCluster"
metadata:
  name: my-nats-cluster
  namespace: nats-stuff
spec:
  size: 5
  version: "1.3.0"
  service:
    maxPayload: 20971520
  pod:
    volumeMounts:
      - name: authconfig
        mountPath: /authconfig
  template: 
    serviceAccountName: my-secrets-getting-sa
    initContainers:
      - name: secret-getter
        image: "...."
        args: ["/authconfig/auth.json"]
        volumeMounts:
          - name: authconfig
            mountPath: /authconfig
    volumes:
      - name: authconfig
        emptyDir: {}
  auth:
    clientAuthFile: "/authconfig/auth.json"
```

wallyqs: That looks good to me 💯

wallyqs: The includes approach under auth section I think will work well, I should mention though that there was a bug in NATS 1.4.1 and earlier with `include` being ignored after being generated by the operator so I had to do this... 🙈 

https://github.com/nats-io/nats-operator/blob/bbff7978648d4411c157c16bded3df51a9a017cf/pkg/util/kubernetes/kubernetes.go#L428-L432

If you are ok with it you could use one of the edge images where this is fixed to avoid this for now, we should be releasing a new image with the fixes some time this month:

```yaml
---
kind: "NatsCluster"
metadata:
  name: "nats-custom"
spec:
  size: 3
  serverImage: "synadia/nats-server"
  version: "edge-2.0.0-RC5"

  pod:
    enableConfigReload: true

  natsConfig:
    debug: true
    trace: true
    maxPayload: 5242880
```

schancel: This should probably be updated to be ReplaceAll then?
```
bytes.Replace(rawConfig, []byte(`"include":`), []byte("include "), 1)
```

schancel: Re: https://github.com/nats-io/gnatsd/commit/441d865c1cd34b2f0ab9750e7286d593be132a7c

Making include a special map keyword at the parser level might bite you in the future.

pkg/cluster/cluster.go
```diff
@@ -188,6 +190,14 @@ func (c *Cluster) checkClientAuthUpdate() error {
 			c.cluster.SetClientAuthSecretResourceVersion(result.ResourceVersion)
 			return c.updateConfigSecret()
 		}
+	} else if c.cluster.Spec.Auth.ClientsAuthFile != "" {
+		hasher := sha1.New()
```
schancel: So it seems like, from the conversation in https://github.com/nats-io/nats-operator/issues/160, that these changes are unnecessary and if the file path is changed, the config update will happen due to `updateCluser`?

wallyqs: that's right, if the file path changes via the CRD update then that would trigger the reload as well

pkg/cluster/cluster.go
wallyqs: right, since there could be more than one include then should change to `ReplaceAll` now. Currently `include` is the only keyword, and we haven't released this change yet... so still time to change it. I added it to try to make the format more compatible with JSON.  One issue from this for example is that since now `include` is a keyword from a map, then can have only a single `include` even though the NATS config allows multiple includes.

schancel:
Okay, if this looks good, I'll try to write some tests.  Looks pretty gnarly though.

schancel:
Also, should I put the sample NatsCluster file somewhere?

wallyqs:
Thanks @schancel ! fwiw I think this approach might better than the other one of injecting all the credentials into the same config secret. The test could be a variation of this one maybe: https://github.com/nats-io/nats-operator/blob/master/test/e2e/config_reload_test.go#L100
It would be great to have an example here to include in the release notes too: https://github.com/nats-io/nats-operator/tree/master/example

test/e2e/config_reload_test.go
```diff
@@ -231,6 +231,164 @@ func TestConfigReloadOnClientAuthSecretChange(t *testing.T) {
...
+	// Make sure we cleanup the secret after we're done testing.
+	defer func() {
+		if err = f.DeleteSecret(cas); err != nil {
+			t.Error(err)
+		}
+	}()
+
+	// Create a NatsCluster resource with a single member, having configuration reloading enabled and using the secret above for client authentication.
+	natsCluster, err = f.CreateCluster(f.Namespace, "test-nats-", size, version, func(natsCluster *natsv1alpha2.NatsCluster) {
+		natsCluster.Spec.Auth = &natsv1alpha2.AuthConfig{
```
schancel: Only thing I changed in the test is this.    Do you want me to roll them together?

Also, how can I run these guys?

wallyqs: There are some notes in here to run the tests https://github.com/nats-io/nats-operator/blob/master/docs/development/development.md#testing

You can also run a single one by doing something like:

```sh
cd $GOPATH/src/github.com/nats-io/nats-operator/test/e2e 
go test ./... -run TestConfigReloadOnNatsServiceRoleUpdates -v --kubeconfig=$HOME/.kube/config
```

Btw I'm also looking at doing a release later today so if it is ok with you I can try to finish the test and push a new images for the operator + reloader sidecar that include the changes.

schancel:
@wallyqs Well, I finally got it compiling.  I still can't setup the e2e tests locally due to virtualization being disabled on the hardware I'm using.  

Do you have any idea what's going on here?

schancel:
Okay it's failing on:

```
	// Wait for the single pod to be created.
	ctx1, fn := context.WithTimeout(context.Background(), waitTimeout)
	defer fn()
	if err = f.WaitUntilFullMeshWithVersion(ctx1, natsCluster, size, version); err != nil {
		t.Fatal(err)
	}
```

Specifically for my test.  There must be something wrong with the manifest being provided to kubernetes.

example/example-nats-cluster-authfile.yaml
```diff
@@ -0,0 +1,32 @@
+# This is an example NatsCluster manifest which uses a 3rd party initContainer
```
wallyqs: I built a new version of the reloader and the following works for me:

```yaml
# This is an example NatsCluster manifest which uses a 3rd party initContainer
# to fetch the authorization credentials from outside kubernetes.
#
# An example of this could be consul-template getting user/passwords from vault.
kind: "NatsCluster"
metadata:
  name: nats-auth-file-example
  namespace: default
spec:
  size: 1
  version: "1.4.1"

  natsConfig:
    maxPayload: 20971520

  pod:
    enableConfigReload: true
    reloaderImage: "wallyqs/nats-server-config-reloader"
    reloaderImageTag: "0.4.4-v1alpha2"
    reloaderImagePullPolicy: "IfNotPresent"

    volumeMounts:
      - name: authconfig
        mountPath: /etc/nats-config/authconfig

  auth:
    clientsAuthFile: "authconfig/auth.json"

  template:
    spec:
      initContainers:
        - name: secret-getter
          image: "busybox"
          command: ["sh", "-c", "echo 'users = [ { user: 'foo', pass: 'bar' } ]' > /etc/nats-config/authconfig/auth.json"]
          volumeMounts:
            - name: authconfig
              mountPath: /etc/nats-config/authconfig
      volumes:
        - name: authconfig
          emptyDir: {}
```

test/e2e/config_reload_test.go
```diff
@@ -101,6 +102,68 @@ func TestConfigReloadOnClientAuthSecretChange(t *testing.T) {
...
+
+	ConfigReloadTestHelper(t, func(natsCluster *natsv1alpha2.NatsCluster, cas *v1.Secret) {
+		natsCluster.Spec.Auth = &natsv1alpha2.AuthConfig{
+			// Use the secret created above for client authentication.
+			ClientsAuthFile: "/authconfig/auth.json",
+		}
+		natsCluster.Spec.Pod = &natsv1alpha2.PodPolicy{
+			// Enable configuration reloading.
+			EnableConfigReload: true,
+			VolumeMounts: []v1.VolumeMount{
```
wallyqs: The reloader image does not get build along with the test so currently have to specify the new reloader image that can be used, for example using these that I pushed:

```
    reloaderImage: "wallyqs/nats-server-config-reloader"
    reloaderImageTag: "0.4.4-v1alpha2"
```

test/e2e/config_reload_test.go
```diff
@@ -101,6 +102,68 @@ func TestConfigReloadOnClientAuthSecretChange(t *testing.T) {
...
+// Then, the test creates a NatsCluster resource that uses this secret for authentication, and makes sure that "user-1" can connect to the NATS cluster.
+// Finally, it removes the entry that corresponds to "user-1" from the authentication secret, and makes sure that "user-1" cannot connect to the NATS cluster anymore.
+func TestConfigReloadOnClientAuthFileChange(t *testing.T) {
+	// Skip the test if "ShareProcessNamespace" is not enabled.
+	f.Require(t, framework.ShareProcessNamespace)
+
+	ConfigReloadTestHelper(t, func(natsCluster *natsv1alpha2.NatsCluster, cas *v1.Secret) {
+		natsCluster.Spec.Auth = &natsv1alpha2.AuthConfig{
+			// Use the secret created above for client authentication.
+			ClientsAuthFile: "/authconfig/auth.json",
```
wallyqs: For the `include` to work it has to be mounted at `/etc/nats-config/authconfig/` since the includes have to be relative to where the main config is located which is `/etc/nats-config/`.

pkg/util/kubernetes/pod.go
```diff
@@ -93,6 +93,10 @@ func natsPodReloaderContainer(image, tag, pullPolicy string) v1.Container {
 			constants.PidFilePath,
 		},
 	}
+	if authFilePath != "" {
+		container.Command = append(container.Command, "-authfile", authFilePath)
```
wallyqs: I might have been wrong on this one... it looks like the reloader would already follow the events in the folder so if the volume is mounted underneath the main nats config file, it would signal as well in case of changes:

https://github.com/nats-io/nats-operator/blob/master/pkg/reloader/reloader.go#L94-L99

pkg/reloader/reloader.go
```diff
@@ -98,6 +99,13 @@ func (r *Reloader) Run(ctx context.Context) error {
 		return err
 	}
 
+	// If we have an auth file configuration, watch it also.
+	if r.AuthFile != "" {
+		if err := configWatcher.Add(r.AuthFile); err != nil {
+			return err
+		}
+	}
```
wallyqs: actually yes this is needed because it takes a digest of the file contents, and only sends signal when it is different...

schancel:
@wallyqs Yeah, I noticed that.  I'm updating the reloader now to be a bit more generic

test/e2e/config_reload_test.go
schancel: :scream: 

That's a huge caveat to add to the documentation.  Will definitely cause confusion.

pkg/util/kubernetes/pod.go
schancel: Also, it's not supposed to be recursively watching.  (According to the documentation)

I created another PR to fix these problems with reloader independently.

wallyqs: That's right... it only works per folder, it is not recursive so would need to list each of the folders that should be tracked for changes.

test/e2e/config_reload_test.go
wallyqs: Agree... we'd have to cover with documentation for now.

wallyqs:
Once the `multi-reloader` branch is merged and this branch is rebased, and updating the reloader command, applying the following to the branch makes things work for me: https://github.com/wallyqs/nats-operator/commit/35ed43536f50ad3b19ab445468c38f5d7d34426f

There was an issue in the reloader test that when decoding JSON it would also accidentally escape `>`  because by default the Go encoder escapes HTML, only workaround seems to be is to setup manually the encoder to avoid escaping html...

```diff
@@ -211,11 +215,24 @@ func ConfigReloadTestHelper(t *testing.T, customizer NatsClusterCustomizerWSecre
 			},
 		},
 	}
-	// Serialize the object containing authentication data.
-	if d, err = json.Marshal(auth); err != nil {
+	// Serialize the object containing authentication data,
+	// we are using wildcard so need to unescape the HTML
+	// which the JSON encoder does by default...
+	buf := &bytes.Buffer{}
+	encoder := json.NewEncoder(buf)
+	encoder.SetEscapeHTML(false)
+	err = encoder.Encode(auth)
+	if err != nil {
+		t.Fatal(err)
+	}
+	buf2 := &bytes.Buffer{}
+	err = json.Indent(buf2, buf.Bytes(), "", "  ")
+	if err != nil {
 		t.Fatal(err)
 	}
+
 	// Create a secret containing authentication data.
+	d = buf2.Bytes()
 	if cas, err = f.CreateSecret(f.Namespace, "data", d); err != nil {
 		t.Fatal(err)
 	}
```

wallyqs:
Merged via https://github.com/nats-io/nats-operator/commit/7ad614fb697fcaaa0b510f97de3f1d412ba3ac3c

PR #163, opened 2019-04-09
Add few missing watch RBAC permissions

wallyqs:
Adds some missing perms that were only applied to the helm chart: https://github.com/nats-io/nats-operator/pull/148/files


wallyqs:
Missed that skaffold needed its own RBAC setup being maintained...,  ok so we have now same RBAC defined in 3 different places in the repo...

pires:
Created #165 for tracking improvements to this.

PR #171, opened 2019-04-13
[reloader] Add support for multiple config files

schancel:
Currently the config file expects there to only be one config file
watched.  However, the nats config file can include other paths, and
we may also want to watch those.

This commit implements being able to specify an arbtirary number of
configuration files to be watched.

### DIFF
--- a/cmd/reloader/main.go
+++ b/cmd/reloader/main.go
@@ -7,11 +7,25 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
-	"github.com/nats-io/nats-operator/pkg/reloader"
+	natsreloader "github.com/nats-io/nats-operator/pkg/reloader"
 	"github.com/nats-io/nats-operator/version"
 )
+
+// StringSet is a wrapper for []string to allow using it with the flags package.
+type StringSet []string
+
+func (s *StringSet) String() string {
+	return strings.Join([]string(*s), ", ")
+}
+
+// Set appends the value provided to the list of strings.
+func (s *StringSet) Set(val string) error {
+	*s = append(*s, val)
+	return nil
+}
 
 func main() {
 	fs := flag.NewFlagSet("nats-server-config-reloader", flag.ExitOnError)
@@ -25,21 +39,28 @@ func main() {
 	var (
 		showHelp    bool
 		showVersion bool
+		fileSet     StringSet
 	)
+
+	nconfig := &natsreloader.Config{}
 	fs.BoolVar(&showHelp, "h", false, "Show help")
 	fs.BoolVar(&showHelp, "help", false, "Show help")
 	fs.BoolVar(&showVersion, "v", false, "Show version")
 	fs.BoolVar(&showVersion, "version", false, "Show version")
 
-	nconfig := &natsreloader.Config{}
 	fs.StringVar(&nconfig.PidFile, "P", "/var/run/nats/gnatsd.pid", "NATS Server Pid File")
 	fs.StringVar(&nconfig.PidFile, "pid", "/var/run/nats/gnatsd.pid", "NATS Server Pid File")
-	fs.StringVar(&nconfig.ConfigFile, "c", "/etc/nats/gnatsd.conf", "NATS Server Config File")
-	fs.StringVar(&nconfig.ConfigFile, "config", "/etc/nats/gnatsd.conf", "NATS Server Config File")
+	fs.Var(&fileSet, "c", "NATS Server Config File (may be repeated to specify more than one)")
+	fs.Var(&fileSet, "config", "NATS Server Config File (may be repeated to specify more than one)")
 	fs.IntVar(&nconfig.MaxRetries, "max-retries", 5, "Max attempts to trigger reload")
 	fs.IntVar(&nconfig.RetryWaitSecs, "retry-wait-secs", 2, "Time to back off when reloading fails before retrying")
 
 	fs.Parse(os.Args[1:])
+
+	nconfig.ConfigFiles = fileSet
+	if len(fileSet) == 0 {
+		nconfig.ConfigFiles = []string{"/etc/nats-config/gnatsd.conf"}
+	}
 
 	switch {
 	case showHelp:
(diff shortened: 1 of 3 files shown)


cmd/reloader/main.go
```diff
@@ -7,12 +7,26 @@ import (
...
+func (s StringSet) String() string {
+	return strings.Join([]string(s), ", ")
+}
+
+// Set appends the value provided to the list of strings.
+func (s StringSet) Set(val string) error {
+	s = append(s, val)
+	return nil
+}
+
```
wallyqs: This is not working for me as it is...

```console
go run cmd/reloader/main.go -c /tmp/the-nats-conf/nats.conf -pid /tmp/nats.pid
2019/04/13 11:47:50 Starting NATS Server Reloader v0.4.4-v1alpha2+git
Error: lstat /etc/nats: no such file or directory
exit status 1
```

Looks like they have to be pointer receivers for it to work:

```
Or you can create custom flags that satisfy the Value interface (with pointer receivers) and couple them to flag parsing by

flag.Var(&flagVal, "name", "help message for flagname")
```

I changed this into the following and now works:

```go
// StringSet is a wrapper for []string to allow using it with the flags package.
type StringSet []string

func (s *StringSet) String() string {
	return strings.Join([]string(*s), ", ")
}

// Set appends the value provided to the list of strings.
func (s *StringSet) Set(val string) error {
	*s = append(*s, val)
	return nil
}
```

```
go run cmd/reloader/main.go -c /tmp/the-nats-conf/nats.conf -c /tmp/the-nats-conf/authconfig/auth.conf -pid /tmp/nats.pid
[/tmp/the-nats-conf/nats.conf /tmp/the-nats-conf/authconfig/auth.conf]
2019/04/13 13:29:29 Starting NATS Server Reloader v0.4.4-v1alpha2+git
2019/04/13 13:29:31 Event: "/tmp/the-nats-conf/.#nats.conf": CREATE 
2019/04/13 13:29:32 Event: "/tmp/the-nats-conf/nats.conf": CHMOD 
2019/04/13 13:29:32 Event: "/tmp/the-nats-conf/nats.conf": WRITE 
2019/04/13 13:29:32 Sending signal to server to reload configuration
2019/04/13 13:29:45 Event: "/tmp/the-nats-conf/authconfig/.#auth.conf": CREATE 
2019/04/13 13:29:47 Event: "/tmp/the-nats-conf/authconfig/auth.conf": CHMOD 
2019/04/13 13:29:47 Event: "/tmp/the-nats-conf/authconfig/auth.conf": WRITE 
```

pkg/reloader/reloader.go
```diff
@@ -110,25 +115,43 @@ func (r *Reloader) Run(ctx context.Context) error {
...
-					continue
+
+			for _, configFile := range r.ConfigFiles {
+				configInfo, err := os.Stat(configFile)
+				if err != nil {
+					log.Printf("Error: %s\n", err)
+					continue WaitForEvent
+				}
+				if !os.SameFile(touchedInfo, configInfo) {
+					continue WaitForEvent
```
wallyqs: I think here need to just `continue` to be able to check the next cached file:
```suggestion
					continue
```

wallyqs:
Sent PR to the branch with a couple of fixes: https://github.com/schancel/nats-operator/compare/multi-reloader...wallyqs:multi-reloader?expand=1

cmd/reloader/main.go
schancel: Derp yeah, obviously.  My bad.

pkg/reloader/reloader.go
```diff
@@ -94,41 +90,63 @@ func (r *Reloader) Run(ctx context.Context) error {
 	// Follow configuration updates in the directory where
 	// the config file is located and trigger reload when
 	// it is either recreated or written into.
-	if err := configWatcher.Add(path.Dir(r.ConfigFile)); err != nil {
-		return err
+	for i := range r.ConfigFiles {
+		// Ensure our paths are canonical
+		r.ConfigFiles[i], _ = filepath.Abs(r.ConfigFiles[i])
+		if err := configWatcher.Add(filepath.Dir(r.ConfigFiles[i])); err != nil {
```
schancel: I was going to ask... Is this the directory because of the way k8s refreshes the files?  I assume it unmounts and remounts the entire dir?

schancel: If that's the case, there should probably be a comment for posterity

cmd/reloader/main.go
```diff
@@ -25,22 +39,29 @@ func main() {
...
+	fs.Var(&fileSet, "c", "NATS Server Config File (may be repeated to specify more than one)")
+	fs.Var(&fileSet, "config", "NATS Server Config File (may be repeated to specify more than one)")
 	fs.IntVar(&nconfig.MaxRetries, "max-retries", 5, "Max attempts to trigger reload")
 	fs.IntVar(&nconfig.RetryWaitSecs, "retry-wait-secs", 2, "Time to back off when reloading fails before retrying")
 
 	fs.Parse(os.Args[1:])
 
+	nconfig.ConfigFiles = fileSet
+	if len(fileSet) == 0 {
+		nconfig.ConfigFiles = []string{"/etc/nats/gnatsd.conf"}
```
schancel: Do you want me to change this to `/etc/nats-config/gnatsd.conf`. I didn't want to change behavior, but I doubt anyone is using that?  It's always set explicitly to the above by the operator it seems.

wallyqs: sounds good to me to change to what actually is being used, thanks

pkg/reloader/reloader.go
wallyqs: yes that’s right too... couldn’t make it work well in k8s unless using the directory

schancel:
Can do tomorrow.

schancel:
Pushed up a commit to figure out why e2e test blew up... :/

wallyqs:
Picked commit https://github.com/nats-io/nats-operator/commit/38a67d92a82f61a2c3b3e3241e3649869be3482a from the branch and merged. Changes to the reloader should not have affected the results from the e2e tests, so taking a look at fixing the flapping test at https://github.com/nats-io/nats-operator/pull/178

PR #174, opened 2019-04-17
Added options to overwrite default ca, key and cert names in TLS secrets

schantaraud:
New options:
```
tls:
    serverSecretCAFileName: ca.crt
    serverSecretCertFileName: tls.crt
    serverSecretKeyFileName: tls.key
    routesSecretCAFileName: ca.crt
    routesSecretCertFileName: tls.crt
    routesSecretKeyFileName: tls.key
```

Use case: kubernetes.io/tls secrets (e.g.: cert-manager-managed certificates)

### DIFF
--- a/pkg/util/kubernetes/kubernetes.go
+++ b/pkg/util/kubernetes/kubernetes.go
@@ -156,9 +156,9 @@ func addTLSConfig(sconfig *natsconf.ServerConfig, cs v1alpha2.ClusterSpec) {
 
 	if cs.TLS.ServerSecret != "" {
 		sconfig.TLS = &natsconf.TLSConfig{
-			CAFile:   constants.ServerCAFilePath,
-			CertFile: constants.ServerCertFilePath,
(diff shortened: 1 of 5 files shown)


pkg/util/kubernetes/kubernetes.go
```diff
@@ -156,9 +156,9 @@ func addTLSConfig(sconfig *natsconf.ServerConfig, cs v1alpha2.ClusterSpec) {
 
 	if cs.TLS.ServerSecret != "" {
 		sconfig.TLS = &natsconf.TLSConfig{
-			CAFile:   constants.ServerCAFilePath,
-			CertFile: constants.ServerCertFilePath,
-			KeyFile:  constants.ServerKeyFilePath,
+			CAFile:   constants.ServerCertsMountPath + "/" + cs.TLS.ServerSecretCAFileName,
```
wallyqs: might be good to start using `filepath.Join(constants.ServerCertsMountPath, cs.TLS.ServerSecretCAFileName)` for these

PR #178, opened 2019-04-27
Use less replicas on upgrade test

wallyqs:
It has been failing a couple of times with timeouts.

### DIFF
--- a/test/e2e/lame_duck_mode_test.go
+++ b/test/e2e/lame_duck_mode_test.go
@@ -33,11 +33,14 @@ type podLDMResult struct {
 	timestamp time.Time
 }
 
-// TestLameDuckModeWhenScalingDown creates a NatsCluster resource with five members and waits for the full mesh to be formed.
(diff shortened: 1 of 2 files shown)


pires:
Well, maybe we could investigate why it is timing out?

wallyqs:
@pires anyway to bubble those errors from the skaffold run?

wallyqs:
It is usually the last test and failing on waiting for the cluster to come up, and since not failing locally guess it might be due to temp lack of resources on the Travis minikube.  We may have some instances being tore down async in between tests that could cause that which could be another thing to look at.

wallyqs:
Locally can make it fail when the test suite takes longer than 10m, so we could extend the `go test` timeout to 20m for example and try to improve the length of some of the tests (always using the same set of nats server images so that they are cached for example).

pires:
OK this will maybe suffice for now.

Are we not using the official container images? Also, container images are cached per Kubernetes node alone.

PR #179, opened 2019-04-28
Enable specifing a `ClientsAuthFile` for user authorization data

wallyqs:
Follow up from https://github.com/nats-io/nats-operator/pull/159 
Includes a few fixes and also extends the tests timeout to 20m since with the tests added in this branch now `go test` takes over 10m which makes the test build fail. /cc @schancel

### DIFF
--- a/example/example-nats-cluster-authfile.yaml
+++ b/example/example-nats-cluster-authfile.yaml
@@ -1,35 +1,40 @@
 # This is an example NatsCluster manifest which uses a 3rd party initContainer
 # to fetch the authorization credentials from outside kubernetes.
 #
-# An example of this could be consul-template getting user/passwords from vault.
 apiVersion: "nats.io/v1alpha2"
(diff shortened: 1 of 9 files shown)


example/example-nats-cluster-authfile.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - name: authconfig
+            mountPath: /etc/nats-config/authconfig
+    volumes:
+      - name: authconfig
+        emptyDir: {}
+  auth:
+    # Needs to be under /etc/nats-config where nats looks
+    # for its config file, or it won't be able to be included
+    # by /etc/nats-config/gnatsd.conf
+    clientAuthFile: "authconfig/auth.json"
```
pires: Shouldn't it be `clientsAuthFile`?

wallyqs: that's right.... thanks

Issue #181, opened 2019-05-07
[Helm] CRD not deleted properly


richerlariviere:
It's been a while since I had the time to play with with the Nats-Operator Helm chart. I noticed that when doing a complete deletion of a chart (`helm delete --purge RELEASE_NAME`):

```shell

# Delete older deployment
$ helm delete --purge nats-operator

# Install chart
$ helm install --name nats-operator charts/nats-operator-0.1.2.tgz --values values/nats-operator.yaml
Error: object is being deleted: customresourcedefinitions.apiextensions.k8s.io "natsserviceroles.nats.io" already exists

# List nats CRDs after a delete operation
$ kube get crd
NAME                                                  CREATED AT
natsclusters.nats.io                                  2019-05-07T11:44:50Z

# Delete natscluster.nats.io CRD
$ kube delete crd/natsclusters.nats.io
customresourcedefinition.apiextensions.k8s.io "natsclusters.nats.io" deleted

# Try to reinstall chart (it works)
$ helm install --name nats-operator charts/nats-operator-0.1.2.tgz --values values/nats-operator.yaml
NAME:   nats-operator
LAST DEPLOYED: Tue May  7 07:57:43 2019
NAMESPACE: default
STATUS: DEPLOYED
... etc ...
```

richerlariviere:
I looked at https://github.com/nats-io/nats-operator/blob/d2009143535e26b1b872befeebe7f81af2da5d52/helm/nats-operator/templates/customresourcedefinition.yaml#L7
And
https://github.com/nats-io/nats-operator/blob/d2009143535e26b1b872befeebe7f81af2da5d52/helm/nats-operator/templates/customresourcedefinition.yaml#L26

The behavior I described in my issue can be explained by this documentation (https://github.com/helm/helm/blob/master/docs/charts_hooks.md#automatically-delete-hook-from-previous-release)

Should we document the fact we need to manually delete CRDs in order to reinstall the chart?

wallyqs:
@richerlariviere adding a note about that sounds good to me too 👍

richerlariviere:
#188

PR #188, opened 2019-05-13
Add docs related to nats.io crd

richerlariviere:
When we want to delete or reinstall NATS-Operator chart, we must ensure that our CRD are uninstalled first. This is because the chart uses helm.sh/hook-delete-policy: before-hook-creation for both its CRD's. Since those CRD's are marked with `scope: Namespaced`, it means it's safe to delete a namespaced instance of NATS-Operator without affecting other instances. People then should be able to understand by themselves why they are getting errors like `Error: object is being deleted: customresourcedefinitions.apiextensions.k8s.io "natsserviceroles.nats.io" already exists`

[See the official Helm documentation](https://github.com/helm/helm/blob/master/docs/charts_hooks.md#automatically-delete-hook-from-previous-release)

### DIFF
--- a/helm/nats-operator/README.md
+++ b/helm/nats-operator/README.md
@@ -40,12 +40,24 @@ parameters that can be configured during installation.
 
 > **Tip**: List all releases using `helm list`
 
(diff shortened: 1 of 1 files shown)


richerlariviere:
Fix #181

PR #195, opened 2019-06-14
Update to use NATS v2 image by default

wallyqs:
Added also handling to use `/nats-server` when using v2.0 images.

### DIFF
--- a/pkg/util/versionCheck/versionCheck.go
+++ b/pkg/util/versionCheck/versionCheck.go
@@ -1,0 +1,29 @@
+package versionCheck
+
+import (
+	"strconv"
+	"strings"
+)
+
+const (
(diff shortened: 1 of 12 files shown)


pkg/util/versionCheck/versionCheck.go
```diff
@@ -0,0 +1,29 @@
...
+	OldNatsBinaryPath = "/gnatsd"
+
+	// NatsBinaryPath after v2 release.
+	NatsBinaryPath = "/nats-server"
+)
+
+func ServerBinaryPath(version string) string {
+	v := strings.Split(version, ".")
+	if len(v) > 0 {
+		majorVersion, err := strconv.Atoi(v[0])
```
ColinSullivan1: Could this have issues with tagged versions like "2.1.0-RC1"?

PR #201, opened 2019-06-26
Change service for clients into headless

wallyqs:
This changes both services created by the operator for a cluster into being headless, and uses the name of the headless service as a seed in the routes configuration in order to create the cluster. This seems to work ok and not cause partitions. For example a cluster like the following:

```yaml
kind: "NatsCluster"
metadata:
  name: "nats"
spec:
  size: 3
  version: "2.0.0"
```

would result into the following configuration:

```js
{
  "port": 4222,
  "http_port": 8222,
  "cluster": {
    "port": 6222,
    "routes": [
      "nats://nats:6222"
    ]
  }
}
```

This also changes the cluster advertise name that is announced by a pod, so if using TLS for the routes now the certificate has to be a wildcard certificate like: `*.<svc-name>` without the `-mgmt` previous to this change (`nats-route://nats-1.nats.default.svc:6222`)

```
nc nats 6222
INFO {"server_id":"NBW7BH2XN6ELMBUFNNVDPDQQ2Q6VYMLHZFXLBTCVTFARY4CMFPZVXSF3","version":"2.0.0","proto":2,"go":"go1.11.10","host":"nats-1.nats.default.svc","port":6222,"max_payload":1048576,"ip":"nats-route://nats-1.nats.default.svc:6222/","nonce":"Ye33jdjhlNBDNlQ","connect_urls":["10.60.0.11:4222"]} 
```

### DIFF
--- a/test/e2e/basic_test.go
+++ b/test/e2e/basic_test.go
@@ -18,6 +18,7 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
@@ -33,7 +34,7 @@ import (
 func TestCreateCluster(t *testing.T) {
 	var (
 		size    = 3
-		version = "1.3.0"
+		version = "1.4.0"
 	)
 
 	var (
@@ -98,7 +99,7 @@ func TestPauseControl(t *testing.T) {
 	var (
 		initialSize = 3
 		finalSize   = 5
-		version     = "1.3.0"
+		version     = "2.0.0"
 	)
 
 	var (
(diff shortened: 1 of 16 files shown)


test/e2e/basic_test.go
```diff
@@ -399,6 +400,7 @@ func TestCreateAndDeleteClusterDependencies(t *testing.T) {
 
 		// Wait for a single pod to be created.
 		pods, err := f.PodsForNatsCluster(natsCluster)
+		fmt.Println("FOUND! ", pods)
```
ColinSullivan1: needed?

test/e2e/basic_test.go
```diff
@@ -408,6 +410,7 @@ func TestCreateAndDeleteClusterDependencies(t *testing.T) {
 
 		// Confirm that there is a service for the NatsCluster.
 		svcs, err := f.ServicesForNatsCluster(natsCluster)
+		fmt.Println("FOUND: ", svcs)
```
ColinSullivan1: needed?

PR #204, opened 2019-07-10
Change the restart policy to pod to always

wallyqs:
By default, in case of failure and nats server is terminated due to a liveness check fail for example we should let kubectl restart it.


ColinSullivan1:
Per our slack discussion, `OnFailure` might be better for upgrades, etc.

PR #217, opened 2019-10-19
Switch to go modules

ekarlso:


wallyqs:
LGTM! Fixes for `go test` applied via https://github.com/nats-io/nats-operator/pull/218

PR #224, opened 2019-11-15
gh-223: updated dockerfile to use go version matching travis and go modules

### LINKED ISSUE #223: Dockerfile never updated to use go modules
Tried building a docker image for ARM and discovered that the dockerfile for the operator was never updated to use go 1.13 or go modules.

mjpitz:
Resolves #223

### DIFF
--- a/docker/bootconfig/Dockerfile
+++ b/docker/bootconfig/Dockerfile
@@ -1,9 +1,8 @@
-FROM golang:1.11-alpine3.8 AS builder
+FROM golang:1.13-alpine3.10 AS builder
 WORKDIR $GOPATH/src/github.com/nats-io/nats-operator/
 RUN apk add --update git
-RUN go get -u github.com/golang/dep/cmd/dep
 COPY . .
-RUN dep ensure -v -vendor-only
+RUN go get ./...
 RUN CGO_ENABLED=0 go build -o /nats-pod-bootconfig ./cmd/bootconfig/main.go
 
 FROM alpine:3.8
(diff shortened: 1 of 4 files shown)


docker/bootconfig/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM golang:1.11-alpine3.8 AS builder
+FROM golang:1.13-alpine3.10 AS builder
 WORKDIR $GOPATH/src/github.com/nats-io/nats-operator/
 RUN apk add --update git
-RUN go get -u github.com/golang/dep/cmd/dep
 COPY . .
-RUN dep ensure -v -vendor-only
+RUN go get ./...
 RUN CGO_ENABLED=0 go build -o /nats-pod-bootconfig ./cmd/bootconfig/main.go
```
wallyqs: Thanks for the PR, also adding `GO111MODULE=on` here I think worked for me but would be better to change to Go 1.13 and since there it is the default

mjpitz: Agreed. I figured I'd bring things up to match the ci/cd job.

mjpitz: TIL: `go mod download`

mjpitz:
Alright, I'm kinda out of ideas on how to fix the ci/cd job. Going to take another look with fresh eyes tomorrow.

wallyqs:
Seems that test has been flapping recently, I'll merge as is and take a look separately.  Thanks a lot for the contribution.

PR #228, opened 2019-11-15
change updateStrategy to Strategy

prune998:
`updateStrategy` is only for Statefulsets, not Deployments

PR #230, opened 2019-11-16
gh-229: specify the proper update strategy on the operator deployment

### LINKED ISSUE #229: Helm chart doesn't specify update strategy correctly
The current helm chart doesn't specify the update strategy correctly and causes any apply or create to fail.

mjpitz:
Resolves #229

### DIFF
--- a/helm/nats-operator/templates/deployment.yaml
+++ b/helm/nats-operator/templates/deployment.yaml
@@ -9,6 +9,12 @@ metadata:
     heritage: {{ .Release.Service }}
 spec:
   replicas: {{ .Values.replicaCount }}
+  strategy:
+    type: {{ .Values.updateStrategy }}
+    {{- if .Values.rollingUpdatePartition }}
+    rollingUpdate:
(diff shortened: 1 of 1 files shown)


helm/nats-operator/templates/deployment.yaml
```diff
@@ -26,12 +32,6 @@ spec:
         {{ .Values.podAnnotations }}
       {{- end }}
     spec:
-      updateStrategy:
-        type: {{ .Values.updateStrategy }}
-        {{- if .Values.rollingUpdatePartition }}
-        rollingUpdate:
-          partition: {{ .Values.rollingUpdatePartition }}
```
mjpitz: I think this is a StatefulSet concept and can be removed from the definition. Can't find references for it outside of the StatefulSet documentation.

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions

mjpitz:
Closing in favor of:
https://github.com/nats-io/nats-operator/pull/228/files

PR #241, opened 2020-01-16
Add options for tls cipher and curve preferences

wallyqs:
Example:

```yaml
---
kind: "NatsCluster"
metadata:
  name: "tls-nats"
spec:
  # Number of nodes in the cluster
  size: 3

  tls:
    # Certificates to secure the NATS client connections
    serverSecret: "nats-server-tls"

    # Name of the CA in serverSecret
    serverSecretCAFileName: "ca.crt"

    # Name of the key in serverSecret
    serverSecretKeyFileName: "tls.key"

    # Name of the certificate in serverSecret
    serverSecretCertFileName: "tls.crt"

    cipherSuites:
      - TLS_...

    curvePreferences:
      - example...
```

### DIFF
--- a/pkg/apis/nats/v1alpha2/cluster.go
+++ b/pkg/apis/nats/v1alpha2/cluster.go
@@ -21,7 +21,7 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


ColinSullivan1:
LGTM, but maybe first add a test to sanity check?

PR #253, opened 2020-03-14
make namespace consistent in templates for clusterScoped

shaunc:
The current templates refer to the cluster namespace as:
```
{{- if and .Values.clusterScoped .Values.cluster.namespace }}
  namespace: {{ .Values.cluster.namespace }}
{{- end }}
```
However, `cluster.namespace` was commented out, so the namespace was not getting set correctly in a number of manifests. (Also, it was missing in a few.) I have uncommented `cluster.namespace` and added its default to be `nats-io`, which seems to be required or strongly recommended.


helm/nats-operator/values.yaml
```diff
@@ -105,7 +105,7 @@ cluster:
   name: nats-cluster
 
   ## Choose namespace for cluster deployment if clusterScoped is set to true
-  # namesapce: "some-namespace"
+  namesapce: "nats-io"
```
wallyqs:
```suggestion
  namespace: "nats-io"
```

Starefossen:
This seams incorrect to me, why would I have the operator in the same namespace as my nats cluster?

PR #255, opened 2020-03-24
accommodate new cert manager behavior

matthiashanel:
ca change was done to prevent server auth to show up in extended key usage.
server change was done for completeness sake
route change was done to make sure client cert is included

- signed off by mh@synadia.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,8 @@ spec:
     name: selfsigning
     kind: ClusterIssuer
   commonName: nats-ca
+  usages: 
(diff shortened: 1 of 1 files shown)


wallyqs:
Related: https://github.com/jetstack/cert-manager/issues/2407

egoon:
@wallyqs Maybe you should have updated the certmanager  apiVersion to `cert-manager.io/v1alpha2` (5 occurences) while you were at it

PR #260, opened 2020-04-02
Use correct binary path when entering lame duck mode.

jdolce:
When upgrading from 1.4.1 to 2.x.x entering lame duck mode would fail as it would lookup the binary server path for the desired version and not the currently running version. 

This patch uses the currently running pod version to lookup the binary server path.


Starefossen:
I think we just experienced this when going from v1.4.1 to 2.1.6 using the v0.6.2 version of the operator. Completely removing all of the nats cluster pods did the trick but we did experience 60 seconds of downtime during this period.

PR #262, opened 2020-04-30
Token per namespace

gifanonim:
Add possibility to create `bound-tokens` for `serviceAccounts` in various namespaces.

`NatsServiceRole` still must be created in the same namespace as 
`NatsCluster` is running, but `bound-token` will be created for `ServiceAccount` 
reusorces that can be placed in various namespaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,7 +410,7 @@ nats-user-example-nats-bound-token         Opaque        1         43m
 
 Please note that `NatsServiceRole` must be created in the same namespace as 
 `NatsCluster` is running, but `bound-token` will be created for `ServiceAccount` 
-reusorces that can be placed in various namespaces.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -408,6 +408,10 @@ nats-admin-user-example-nats-bound-token   Opaque        1         43m
 nats-user-example-nats-bound-token         Opaque        1         43m
 ```
 
+Please note that `NatsServiceRole` must be created in the same namespace as 
+`NatsCluster` is running, but `bound-token` will be created for `ServiceAccount` 
+reusorces that can be placed in various namespaces.
```
wallyqs, line 413: Small typo fix here:
```suggestion
resources that can be placed in various namespaces.
```

PR #267, opened 2020-05-19
Propagate Gateways reject_unknown option

philpennock:


pkg/apis/nats/v1alpha2/cluster.go
```diff
@@ -182,6 +182,11 @@ type GatewayConfig struct {
 	// for external access.
 	Port int `json:"hostPort,omitempty"`
 
+	// RejectUnknown indicates to use an implicit authorization map for inbound
+	// connections, consisting of the names of our outbound gateway
+	// declarations.
+	RejectUnknown bool `json:"reject_unknown,omitempty"`
```
wallyqs, line 188: think should make it camel case to be consistent with k8s yaml world here

wallyqs:
Example run:

```
kind: "NatsCluster"
metadata:
  name: "nats-dev"
spec:
  size: 1

  natsConfig:
    debug: true
    trace: true

  pod:
    enableClientsHostPort: true
    advertiseExternalIP: true

  gatewayConfig:
    name: "gw"
    hostPort: 7522
    rejectUnknown: true

    gateways:
      - name: remote-nats
        url: "nats://external-nats-gateway:7522"

  # Required to be able to lookup public ip address from a server.
  template:
    spec:
      serviceAccountName: "nats-server"
```

result:

```json
kubectl get secret nats-dev -o yaml | grep nats.conf | awk '{print $2}' | base64 -D
{
  "port": 4222,
  "http_port": 8222,
  "cluster": {
    "port": 6222,
    "routes": [
      "nats://nats-dev-1.nats-dev-mgmt.default.svc:6222"
    ]
  },
  "debug": true,
  "trace": true,
  "logtime": true,
  include  "advertise/client_advertise.conf",
  "gateway": {
    "name": "gw",
    "port": 7522,
    "gateways": [
      {
        "name": "remote-gw",
        "url": "nats://external-nats-gateway:7522"
      }
    ],
    "reject_unknown": true,
    include  "advertise/gateway_advertise.conf"
  }
}
```

pkg/apis/nats/v1alpha2/cluster.go
wallyqs, line 188:
```suggestion
	RejectUnknown bool `json:"rejectUnknown,omitempty"`
```

pkg/apis/nats/v1alpha2/zz_generated.deepcopy.go
```diff
@@ -481,6 +482,16 @@ func (in *ServiceRoleSpec) DeepCopy() *ServiceRoleSpec {
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *TLSConfig) DeepCopyInto(out *TLSConfig) {
 	*out = *in
+	if in.CipherSuites != nil {
```
kozlovic, line 485: What Go version do you guys have to support? `TLSConfig` has `.Clone()` for quite a bit now, so you may not need custom deep copy.

wallyqs: This is generated code from k8s Ivan so we just apply as is... 🙈 but maybe the code generation tool will eventually support it.

philpennock: Yeah, you can split the PR into the two distinct commits; I did a generate first, committed that to have a baseline; then made my changes, committed, and then generated again, no diff -- my guess being that an inline simple type needed no extra logic.

Starefossen:
Just for future reference; it is very hard for users of the NATS Operator to understand what the changes are when there are no context/descriptions to the pull requests like this and no changelog. We have to resort to digging through the actual changed code each time we upgrade which is very time consuming.

philpennock:
Hrm, I thought the commit message was pretty clear: `Propagate Gateways reject_unknown option` .. the nats-server has options, some are in the gateways section, we want to be able to propagate that option through the operator, so that's what the commit message (and thus PR title) is.  No previously working functionality was broken, nothing needed to be called out to warn folks, we added the ability to set one more NATS config option.  I like detailed commit messages where they make sense, and I write a lot of extremely verbose commit-messages.  But in this case, I could sufficiently describe things more briefly.  Less verbiage leads to easier to parse and scan text.

I guess we could copy/paste documentation about the option into the commit message too?

PR #269, opened 2020-05-22
Improve codegen performance

whynowy:
This change reduces time cost of `make gen` to:

```
real	0m12.943s
user	0m22.032s
sys	0m6.962s
```

It was:

```
real	0m46.666s
user	0m57.515s
sys	0m56.648s
```

### DIFF
--- a/hack/tools.go
+++ b/hack/tools.go
@@ -1,0 +1,16 @@
+// +build tools
+
+// This package contains code generation utilities
+// This package imports things required by build scripts, to force `go mod` to see them as dependencies
+package tools
+
+import (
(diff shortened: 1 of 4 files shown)


hack/tools.go
```diff
@@ -0,0 +1,16 @@
+// +build tools
+
+// This package contains code generation utilities
```
whynowy, line 3: This is a recommended way from `code-generator`, see https://github.com/kubernetes/code-generator/blob/master/tools.go

wallyqs:
Github a bit broken right now, but confirmed the changes locally.

PR #272, opened 2020-06-10
Adds support for setting NATS Pod annotations via Helm Chart

BenConstable:
Support for setting annotations on NATS Pods created by the Operator was added in #111, but these were not able to be set via the Helm Chart.

This PR adds the ability to set annotations on Operator-created Pods though a Chart value. It also tidies up a bit of formatting in the `natscluster.yaml` template, and adds a missing documentation entry for the `cluster.resources` Chart value.


BenConstable:
Nice, thanks!

PR #284, opened 2020-09-04
Add leaf node remotes config

### LINKED ISSUE #280: Create nats leaf remote connection
Hello!

Is there any documentation for how to create the remote connection to a leaf node? 

In my supercluster, I'm creating the cluster with 
```
apiVersion: nats.io/v1alpha2
kind: NatsCluster
metadata:
  labels:
    app: global-nats
  name: global-nats-cluster
spec:
  natsConfig:
    writeDeadline: 5s
    debug: true
    trace: true
  pod:
    enableConfigReload: true
    enableMetrics: true
    metricsImage: synadia/prometheus-nats-exporter
    metricsImageTag: 0.5.0
    reloaderImage: connecteverything/nats-server-config-reloader
    reloaderImagePullPolicy: IfNotPresent
    reloaderImageTag: 0.6.0
  size: 2
  version: 2.1.2
  leafnodeConfig:
    hostPort: 4224
```

I assume there's a way to specify the remote leaf connections in a similar way?

variadico:
Currently, there is no way to specify leaf node remotes. This adds config
support for leaf node remotes and also includes some example YAML to deploy on
Kubernetes.

Fixes #280


example/leafnode-remotes-gateways.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  natsConfig:
+    debug: true
+    trace: true
+
+  pod:
+    volumeMounts:
+    - name: user-credentials
+      mountPath: /etc/nats-creds
+      readOnl: true
```
wallyqs, line 29: Maybe:
```suggestion
      readOnly: true
```

pkg/apis/nats/v1alpha2/cluster.go
```diff
@@ -201,9 +201,16 @@ type RemoteGatewayOpts struct {
 	URL string `json:"url,omitempty"`
 }
 
+// LeafNodeRemote is the URL for remote NATS system.
+type LeafNodeRemote struct {
+	URL         string `json:"url,omitempty"`
```
wallyqs, line 206: I think we should probably also support the array version:
```
URL         []string `json:"urls,omitempty"`
```

example/leafnode-remotes-gateways.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+  pod:
+    volumeMounts:
+    - name: user-credentials
+      mountPath: /etc/nats-creds
+      readOnly: true
+
+  leafnodeConfig:
+    remotes:
+    - url: nats://1.2.3.4:7422
```
wallyqs, line 33: `url` can be many comma separated, but can also be an array `urls: ["nats://...", "nats://..."]` can be common in case to explicit about the clusters

pkg/apis/nats/v1alpha2/cluster.go
variadico: Is this possible NATS config?
```
leafnodes {
    remotes = [
        {
          url: ["nats-leaf://s3cr3t@foo", "nats-leaf://s3cr3t@foo"]
        },
    ]
}
```

wallyqs: yes also possible

wallyqs (approved):
LGTM! Might be good to have support for array version of the remotes too, also we need to add the option for remote leafnode TLS certs and the volume in a similar way as it is done here with the credentials.

pkg/apis/nats/v1alpha2/cluster.go
variadico: Ah, okay. Sounds good. :+1:

variadico:
Okay, so looks like the controller code generator doesn't like if we use `interface{}` types in the YAML. This means we can't overload the type of `url` in YAML.

However, I exposed two keys in the YAML, so both of these are possible. 

```yaml
  leafnodeConfig:
    remotes:
    - url: nats://1.2.3.4:7422
      credentials: /etc/nats-creds/user.ncreds

# and

  leafnodeConfig:
    remotes:
    - urls: [nats://1.2.3.4:7422, nats://1.2.3.4:7423]
      credentials: /etc/nats-creds/user.ncreds

# and technically even this

  leafnodeConfig:
    remotes:
    - url: nats://1.2.3.4:7422
      urls: [nats://1.2.3.4:7422, nats://1.2.3.4:7423]
      credentials: /etc/nats-creds/user.ncreds

# in the above case, all of the URLs get combined into 1 array of URLs
```

All combinations of these URLs keys get merged into this.
```
leafnodes {
    remotes = [
        {
          urls: [nats://1.2.3.4:7422, ...]
        },
    ]
}
```

wallyqs (approved):
LGTM!

Kampe:
Question for you all, when I start my nats servers I should see some connection logs about the leaf nodes connecting?  Is there a way to verify the connections?

variadico:
Yeah, you should see something like this.

```
nats-server -c /tmp/leaf.conf 
....
[3704] 2019/12/09 09:55:31.548308 [INF] Listening for client connections on 127.0.0.1:4111
...
[3704] 2019/12/09 09:55:31.549404 [INF] Connected leafnode to "localhost"
```

Source: https://docs.nats.io/nats-server/configuration/leafnodes

PR #288, opened 2020-10-09
Add websocket support

variadico:
This change exposes a `websocketConfig` block in the `NatsCluster` CRD. This
allows clients to connect to a cluster by using WebSockets. Note: TLS is
required for WebSocket connections.

Also, currently, WebSockets are only supported in NATS Server nightly
images.


pkg/util/versionCheck/versionCheck.go
```diff
@@ -12,9 +12,16 @@ const (
 
 	// NatsBinaryPath after v2 release.
 	NatsBinaryPath = "/nats-server"
+
+	// NightlyNatsBinaryPath is there the server is for nightly images.
+	NightlyNatsBinaryPath = "/bin/nats-server"
 )
 
 func ServerBinaryPath(version string) string {
+	if strings.HasPrefix(version, "nightly") {
```
wallyqs, line 21: 👍

pkg/util/versionCheck/versionCheck.go
```diff
@@ -12,9 +12,16 @@ const (
 
 	// NatsBinaryPath after v2 release.
 	NatsBinaryPath = "/nats-server"
+
+	// NightlyNatsBinaryPath is there the server is for nightly images.
+	NightlyNatsBinaryPath = "/bin/nats-server"
```
wallyqs, line 17: maybe we should fix the nightly images which are made here I think, add an alias instead of adding to the operatoe: https://github.com/nats-io/jetstream/blob/master/Dockerfile

wallyqs (approved):
LGTM

pkg/util/versionCheck/versionCheck.go
wallyqs: ok let's add this back instead

variadico: Readded.

pkg/util/kubernetes/pod.go
```diff
@@ -81,6 +86,16 @@ func natsPodContainer(container v1.Container, clusterName, version string, serve
 		}
 		ports = append(ports, port)
 	}
+	if websocketPort > 0 {
+		port := v1.ContainerPort{
+			Name:          "websocket",
+			ContainerPort: int32(websocketPort),
+			Protocol:      v1.ProtocolTCP,
+			HostPort:      int32(websocketPort),
```
wallyqs, line 94: TODO: Make optional

wallyqs (approved):
LGTM 👾

PR #292, opened 2020-10-23
Provide more diagnostics

philpennock:
Log more details at startup and be a little more explicit later on.
Find out things like "not watching any config files".

Add some jitter to our backoff delays, to avoid phase-locked problems.


wallyqs (approved):
LGTM!!

PR #293, opened 2020-10-30
Allow loading cert information from elsewhere

philpennock:
Rather than require that all the TLS items come from one secret, allow the
filename for each item to be an absolute path instead.

Change the README example to highlight using this to point the CA bundle
elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,13 @@ with the clients.
 $ kubectl create secret generic nats-clients-tls --from-file=ca.pem --from-file=server-key.pem --from-file=server.pem
 ```
 
+Consider though that you may wish to independently manage the certificate
+authorities for routes between clusters, to support the ability to roll
+between CAs or their intermediates.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -208,12 +215,25 @@ spec:
...
     # Name of the certificate in routesSecret
     routesSecretCertFileName: "tls.crt"
+
+  template:
+    spec:
+      containers:
+      - name: "nats"
+        volumeMounts:
+        - name: "ca-bundle"
+          mountPath: "/etc/ca-bundle"
```
wallyqs, line 231: this one would be a single file?

philpennock: Yes.

philpennock: Or rather: there should be a single file inside that directory.

philpennock:
(rebased on top of the release changes)

wallyqs (approved):
LGTM

PR #295, opened 2020-11-02
Merge volumeMounts, rather than replace

philpennock:
If the template spec for the pod already includes any volumeMount items then we
should preserve them, not replace them.


wallyqs (approved):
LGTM

philpennock:
(rebased)

Issue #300, opened 2020-12-15
HostPort when enableClientsHostPort is set to false


mtryfoss:
Hi,

I see that gateway, leaf or websocket config automatically enables listening on a HostPort. In my setup, we use Calico CNI where all Pods do have direct connectivity to our environment. I use discovery through a service with "externalTrafficPolicy: Local" which seems to work perfectly.

	if gatewayPort > 0 {
		port := v1.ContainerPort{
			Name:          "gateway",
			ContainerPort: int32(gatewayPort),
			Protocol:      v1.ProtocolTCP,
			HostPort:      int32(gatewayPort),
		}
		ports = append(ports, port)
	}
	if leafnodePort > 0 {
		port := v1.ContainerPort{
			Name:          "leaf",
			ContainerPort: int32(leafnodePort),
			Protocol:      v1.ProtocolTCP,
			HostPort:      int32(leafnodePort),
		}
		ports = append(ports, port)
	}
	if websocketPort > 0 {
		port := v1.ContainerPort{
			Name:          "websocket",
			ContainerPort: int32(websocketPort),
			Protocol:      v1.ProtocolTCP,
			HostPort:      int32(websocketPort),
		}
		ports = append(ports, port)
	}

Would you accept a PR making it possible to disable this behaviour (enabled by default), making it only request a ContainerPort? 
For example: 

disableGatewaysHostPort
disableLeafnodesHostPort
disableWebsocketsHostPort

wallyqs:
Sounds like a good change to me, we'd gladly accept a PR like that.

mtryfoss:
Updated the PR with new master branch now. Hope it looks ok.

PR #301, opened 2020-12-16
Add parameters to disable HostPorts for gateways, leafnodes and websockets

mtryfoss:
PR related to https://github.com/nats-io/nats-operator/issues/300

Restructured args for natsPodContainer() a bit and introduced a struct of bools to control if HostPorts should be used or not.

I also added a check against AdvertiseExternalIP which ignores these new parameters if that is true. It makes no sense to advertise external IP without using a HostPort?

Tested in my environment with success.


mtryfoss:
I updated the master branch after the other PR. Does this look ok?

wallyqs (approved):
LGTM! Thanks for the contributions

PR #302, opened 2020-12-16
Support for auth on cluster routes

mtryfoss:
After some investigation, I couldn't make authentication for internal cluster routers work. After looking through the code, I couldn't find anything actually producing it though the json structure for config was defined.

Here's my approach for it. I tested basic scenarios and it seems to work.

Will require an extra secret since cluster auth for routes only supporting one username.

NatCluster definition example:

```yaml
  auth:
    clusterAuthSecret: "nats-routes-credentials"
    clusterAuthTimeout: 1
```

```
apiVersion: v1
kind: Secret
metadata:
  name: nats-routes-credentials
data:
  user.ncreds: <base64 here>
```

json-data which should be base64 in the secret above:

```json
{
  "username": "natsclusteruser",
  "password": "verynicepassword"
}
```

It will produce the following config:

```
    "routes": [
      "nats://natsclusteruser:verynicepassword@nats-nats-1.nats-nats1-mgmt.default.svc:6222",
      "nats://natsclusteruser:verynicepassword@nats-nats-2.nats-nats1-mgmt.default.svc:6222",
      "nats://natsclusteruser:verynicepassword@nats-nats-3.nats-nats1-mgmt.default.svc:6222"
    ],
    "authorization": {
      "username": "natsclusteruser",
      "password": "verynicepassword",
      "timeout": 1
    }
```

### DIFF
--- a/pkg/apis/nats/v1alpha2/cluster.go
+++ b/pkg/apis/nats/v1alpha2/cluster.go
@@ -35,6 +35,12 @@ const (
 	// the NATS cluster.
 	clientAuthSecretResourceVersionAnnotationKey = "nats.io/cas"
 
+	// clusterAuthSecretResourceVersionAnnotationKey is the key of
+	// the annotation that holds the last-observed resource
+	// version of the secret containing authentication data for
+	// the NATS cluster.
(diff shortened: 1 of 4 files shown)


wallyqs (approved):
LGTM! Thanks

PR #304, opened 2020-12-19
Fix bugs related to reloading when secrets is changed

mtryfoss:
Fixed a copy-paste bug I introduced in my previous PR. Also fixed an existing issue when the operator was controlling multiple nats cluster in the same namespace sharing the same secret. Only one of them was reloaded when the secret was changed.


mtryfoss:
@wallyqs did you have time to look at this?

PR #311, opened 2021-03-29
Add cluster name configuration to nats.conf

rsafonseca:
This PR passes the cluster name to the actual NATS pods nats.conf
Currently, the pods created by the operator are unaware of their cluster name and generate a dynamic cluster name on startup, which sometimes causes issues when rolling out the nodes, eg. after an upgrade.


variadico (approved):
Seems good. Looks like the files haven't been `gofmt`-ed, but we'll do that on our side to speed up the process.

PR #316, opened 2021-04-16
pkg: add gatewayport and leafnodeport to service

squat:
This commit ensures that the gateway and leafnode ports are added to the
management service created for a cluster.

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>

Plz let me know if these ports should also be added to the client service.
The ports were added to the headless service to match the Helm charts
and raw YAML files, e.g. 
https://github.com/nats-io/k8s/blob/master/helm/charts/nats/templates/service.yaml#L35-L38
Similarly, the port names are pluralized, i.e. `leafnodes` and `gateways`
to match.

### DIFF
--- a/pkg/util/kubernetes/kubernetes.go
+++ b/pkg/util/kubernetes/kubernetes.go
@@ -138,6 +138,8 @@ func CreateMgmtService(
 	clusterName, clusterVersion, ns string,
 	owner metav1.OwnerReference,
 	websocketPort int,
+	gatewayPort int,
+	leafnodePort int,
 ) error {
 	ports := []v1.ServicePort{
 		{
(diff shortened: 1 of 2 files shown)


pkg/util/kubernetes/kubernetes.go
```diff
@@ -138,6 +138,8 @@ func CreateMgmtService(
 	clusterName, clusterVersion, ns string,
 	owner metav1.OwnerReference,
 	websocketPort int,
+	gatewayPort int,
+	leafenodePort int,
```
wallyqs, line 142:
```suggestion
	leafnodePort int,
```

wallyqs:
I think just adding to the management service is fine, thanks for the contribution!

pkg/util/kubernetes/kubernetes.go
squat: :+1:

squat:
@wallyqs do we need to do anything special to get the CI to finish?

wallyqs:
Not sure what changed in travis, the build itself seems fine: https://travis-ci.com/github/nats-io/nats-operator/builds/223332812

PR #333, opened 2021-09-08
Update to k8s 1.17.17

variadico:
This updates our Kubernetes code to version [1.17.17](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.17.md#v11717).

The most interesting update is that CRDs now use the `apiextensions/v1` API, instead of the old `apiextensions/v1beta1` API.


wallyqs (approved):
LGTM!!! 🎉 🎉 🎉

.travis.yml
```diff
@@ -19,10 +19,10 @@ install:
 before_script:
 - go version
 - docker version
-- curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.15.12/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+- curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.17.17/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
```
invidian, line 22: Why 1.17 when it already reached end of life? Why not something more up to date?

variadico: We want to get to latest. But... we originally attempted an update from 1.12 to 1.22 and that didn't go so well. 😅 
We'll try to keep the updates coming.

invidian: Make sense, thanks!

PR #335, opened 2021-09-09
Update to k8s 1.20.10

variadico:
This updates our Kubernetes code to version [1.20.10](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#v12010), EOL 2022-02-28.

The most interesting update is that we had to rewrite lots code to pass a context and options.

I'm also updating the API version used in Skaffold to `skaffold/v2beta21`. Luckily, no code changes were required for that.


wallyqs (approved):
LGTM ⚡

PR #336, opened 2021-09-28
Update to k8s 1.22.2

variadico:
This updates our Kubernetes code to version 1.22.2, EOL 2022-10-28.

In addition, it updates the version of nats-io/nats.go to 1.12.3. Also,
it also updates the version of the NATS Server used in the lame duck
test to 2.6.1 because the 2.0.0 server failed intermittently. Finally,
this changes the minikube start command to use the correct
service-account keys.


wallyqs (approved):
LGTM

PR #338, opened 2022-01-31
added custom signal, refactored loop

siredmar:
This PR allows the config reloader to send a custom signal to the nats server. This can be useful where the user is forced to restart the nats server in order to load a modified configuration (e.g. modified leafnode settings).

Using the config reloader, this PR fixes https://github.com/nats-io/nats-server/issues/1921 as a workaround by allowing the reloader to send a custom signal, e.g. sigterm forcing to terminate the nats server. On the next startup the new configuration will be used.


pkg/reloader/reloader.go
```diff
@@ -195,18 +202,24 @@ WaitForEvent:
...
+			return err
+		}
+
 		// Configuration was updated, try to do reload for a few times
 		// otherwise give up and wait for next event.
 	TryReload:
 		for {
-			log.Println("Sending signal to server to reload configuration")
-			err := r.proc.Signal(syscall.SIGHUP)
+			log.Printf("Sending signal '%s' to server to reload configuration\n", r.Signal.String())
```
philpennock, line 221: nb: `log` package prints include the trailing newline automatically (but at least won't duplicate one already present).

philpennock (approved):
LGTM

pkg/reloader/reloader.go
siredmar: Deleted all newlines from log messages.

wallyqs:
Thanks for the contribution! I will merge as is since LGTM, but actually the main source of the reloader process was moved to be located at: https://github.com/nats-io/nack/blob/main/cmd/nats-server-config-reloader/main.go so I have ported over your changes as part of this PR: https://github.com/nats-io/nack/pull/62

Issue #339, opened 2022-02-10
tolerate-unready-endpoints annotation has been deprecated


aojea:
This annotation was deprecated in 1.13 and is not used in EndpointSlices, that is an evolution of the Endpoint controller.

The project should use the corresponding service API field instead of the annotation

https://github.com/nats-io/nats-operator/blob/861edeaa076c4b91bda3ee853c6e77a358c22de7/pkg/util/kubernetes/kubernetes.go#L883-L884

xref: https://github.com/kubernetes/kubernetes/pull/108020

PR #341, opened 2022-02-11
don't use deprecated TolerateUnreadyEndpointsAnnotation

aojea:
The annotation was deprecated in Kubernetes 1.13, will not be
longer available since 1.24 and only works for the Endpoints
controller.

Use the corresponding Service API field, that is also compatible
with EndpointSlices.

Fixes: https://github.com/nats-io/nats-operator/issues/339


wallyqs (approved):
LGTM, thanks

Issue #88, opened 2018-10-31
Nats-Operator incompatible with istio?


therealmitchconnors:
When I follow the instructions in the project readme to create a nats cluster with 3 members on a gke cluster using istio, all three members immediately show unhealthy and quickly go to crashloopbackoff.  Is there something additional I need to do to get nats-operator to play nice with a service mesh?

My Nats Cluster:
```
echo '
apiVersion: "nats.io/v1alpha2"
kind: "NatsCluster"
metadata:
  name: "example-nats-cluster"
spec:
  size: 3
  version: "1.3.0"
' | kubectl apply -f -
```
Log from one member:
```
[1] 2018/10/30 20:27:15.907885 [INF] Starting nats-server version 1.3.0
[1] 2018/10/30 20:27:15.907943 [INF] Git commit [eed4fbc]
[1] 2018/10/30 20:27:15.908133 [INF] Starting http monitor on 0.0.0.0:8222
[1] 2018/10/30 20:27:15.908194 [INF] Listening for client connections on 0.0.0.0:4222
[1] 2018/10/30 20:27:15.908208 [INF] Server is ready
[1] 2018/10/30 20:27:15.908541 [INF] Listening for route connections on 0.0.0.0:6222
[1] 2018/10/30 20:27:15.914868 [ERR] Error trying to connect to route: dial tcp 10.12.12.4:6222: connect: connection refused
[1] 2018/10/30 20:27:16.930604 [ERR] Error trying to connect to route: dial tcp 10.12.12.4:6222: connect: connection refused
[1] 2018/10/30 20:27:17.935214 [INF] 10.12.12.4:6222 - rid:1 - Route connection created
[1] 2018/10/30 20:27:17.940613 [INF] 127.0.0.1:41486 - rid:2 - Route connection created
[1] 2018/10/30 20:27:18.962862 [INF] 10.12.12.4:6222 - rid:3 - Route connection created
```
(and the Route connection messages continue 290 times before the container is shut down as unhealthy)

My Istio deployment is the default Isitio App from the GCP marketplace, with three nodes in it.  
K8S version info:
```
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.7", GitCommit:"0c38c362511b20a098d7cd855f1314dad92c2780", GitTreeState:"clean", BuildDate:"2018-08-20T10:09:03Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #346, opened 2022-08-29
change service port name to support Istio explicit protocol selection

nberlee:
Kubernetes has a native construct for this called `appProtocol`. Kubernetes
expects either a IANA protocol or a custom protocol prefix with a domain.

I tried setting `appProtocol` to `nats.io/client` but that does not effect
anything Istio, as istio expects only these protocols:
https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

Setting `appProtocol` to `tcp` probably does work, but tcp is not a valid
iana service name. See:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

Therefor the least intrusive, compatible leaves us with prefixing the service
port name.

Fixes: #88

### DIFF
--- a/pkg/util/kubernetes/kubernetes.go
+++ b/pkg/util/kubernetes/kubernetes.go
@@ -112,7 +112,7 @@ func CreateClientService(
 ) error {
 	ports := []v1.ServicePort{
 		{
-			Name:       "client",
+			Name:       "tcp-client",
 			Port:       constants.ClientPort,
 			TargetPort: intstr.FromInt(constants.ClientPort),
 			Protocol:   v1.ProtocolTCP,


vitalii-buchyn-exa:
hi, sorry for postmortem post, but is this fix needed for tcp/6222 server port as well?
thank you

nberlee:
Typically no...  As running NATS inside the istio-mesh may break other things also... but it could not hurt

If I understand correctly 6222 is only used from server to server. And I think most people will run NATS outside the istio-mesh as you run into other problems as well when you run it inside the mesh.

As 4222 can be called from a pod inside the istio-mesh, you need the explicit protocol selection.

Issue #126, opened 2019-02-15
Do not lock to `nats-io` namespace when using `ClusterScoped`


prune998:
I don't understand why we force the use of the `nats-io` namespace here.
It's a huge limitation only to foolproof people trying to deploy many nats-operators in different namespaces !

At least, add an option, like `feature-gates`, to allow to specify the Namespace to use for leader election.

My point of view here is that people starting many nats-operators in different NS will end with troubles, which is fine (and documented).

Simply remove code at :
https://github.com/nats-io/nats-operator/blob/33a51993af2ad3a85338ab752961d5f0a4ac45b3/cmd/operator/main.go#L108

Or maybe we should find a better foolproof solution to manage the leader election ? If yes we can start the discussion and I'll be happy to add and PR that.

wallyqs:
Thanks for the feedback, I think an option to be able to specify the namespace for leader election would be a good addition.

pires:
> I don't understand why we force the use of the `nats-io` namespace here.

Using a fixed, well-known namespace - as opposed to allowing for it to be specified using a flag, as you suggest - is about the only way we have to guarantee that there is a single instance of `nats-operator` operating cluster-wide.
Should we allow for the namespace to be specified using a flag or some other method, there would be nothing preventing users from shooting themselves in the foot because they didn't read the warning sign (documentation).

> It's a huge limitation only to foolproof people trying to deploy many nats-operators in different namespaces !

What makes you think this is _a huge limitation_?

> Or maybe we should find a better foolproof solution to manage the leader election ?

If you find a (better) way then, by all means, please present your point of view for us to discuss it.

That being said, if documentation is enough for you, then there's no need to act _smart_ in many other places of the code, such as:
- https://github.com/nats-io/nats-operator/blob/master/cmd/operator/main.go#L301
- https://github.com/nats-io/nats-operator/blob/master/cmd/operator/main.go#L318
- https://github.com/nats-io/nats-operator/blob/master/cmd/operator/main.go#L339
- https://github.com/nats-io/nats-operator/blob/master/cmd/operator/main.go#L353

prune998:
Actually, the foolproof functions you are mentioning could already do the trick and should be able to detect if either : 
- an operator is already running in the same namespace
- an operator with the `cluster-scope` flag is already running in any namespace

You could also merge the two in a single one that would detect both cases and only refuse to start if : 
[...]

PR #167, opened 2019-04-10
WIP: allow for changing the namespace where operator runs

supriya-premkumar:
Changes:

- Changed KubernetesNamespaceNatsIO to a var.
- This will allow a compile time override of nats namespace for advanced k8s admins to configure a 
   non nats-io namespace for cluster scoped deployment.
- Enables workflows referred by https://github.com/nats-io/nats-operator/issues/126

PR #347, opened 2023-06-16
add feature gate to bypass namespace enforcement

colstrom:
This seems to be the only place where this is used, and the comment suggests it's intended to make things easier. It would be nice if there was a way to opt of out this restriction, so this PR adds a way to do that. It's using the same feature-gate mechanism as the `ClusterScoped` option, since that seemed appropriate, but if there's a better place for it, just let me know.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: focal
 
 language: go
 go:
-  - "1.16.7"
+  - "1.17.13"
 go_import_path: github.com/nats-io/nats-operator
 
 env:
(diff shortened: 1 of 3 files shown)


colstrom:
Looking through the other issues and PRs, this would resolve #126 in a way that makes it explicit that the user is opting out of a safety feature, while still not requiring the namespace to be manually configured. It also supersedes #167 by not making it a build-time decision.

Speaking of build-time... looking at the CI failure above:

```text
../../cloudflare/cfssl/vendor/github.com/mattn/go-sqlite3/sqlite3_opt_serialize.go:41:26: undefined: math.MaxInt
```

It looks like a dependency unrelated to this change is referencing a constant that was first defined in [Go 1.17](https://pkg.go.dev/math@go1.17#pkg-constants), while the build environment is using [Go 1.16.7](https://pkg.go.dev/math@go1.16.7#pkg-constants). I'm not particularly familiar with Travis CI, but I can try to see about bumping the version of Go to 1.17.13 (the latest patch of the closest minor version to what's in use currently), unless there's any objection to that.

